PR #129, opened 2019-12-12
Minor fix in docs for vcr_configure()

aaronwolen:
This had me scratching my head for a while 😄


sckott:
thanks @aaronwolen !

(no worries about failing Travis test - seems broken due to some problems not related to this pkg)

sckott:
you're right, thanks for the fix!

Issue #151, opened 2020-02-11
Cassette options in vcr configuration are ignored


aaronwolen:
The following cassette options can be set w/ `vcr_configure()`

- `record`
- `match_requests_on`
- `re_record_interval`
- `clean_outdated_http_interactions`
- `allow_unused_http_interactions`
- `serialize_with`
- `persist_with`
- `preserve_exact_body_bytes`

but aren't actually used anywhere. It seems like `insert_cassette()` and `use_cassette()` should inherit values for those params from `vcr_c` rather than providing their own defaults. 

Quick example that highlights where the current behavior deviates from my expectation:

``` r
devtools::dev_mode(on = TRUE)
#> Dev mode: ON
library(vcr)

vcr_configure(
  record = "none",
  match_requests_on = "body",
  preserve_exact_body_bytes = TRUE
)
#> <vcr configuration>
#>   Cassette Dir: .
#>   Record: none
#>   URI Parser: crul::url_parse
#>   Match Requests on: body
#>   Preserve Bytes?: TRUE
#>   Logging?: FALSE
#>   ignored hosts: 
#>   ignore localhost?: FALSE
#>   Write disk path:

cas <- insert_cassette("configured-insert")

# expected "none"
cas$record
#> [1] "once"

# expected "body"
cas$match_requests_on
#> [1] "method" "uri"

# expected 
cas$preserve_exact_body_bytes
#> [1] FALSE
```

<sup>Created on 2020-02-11 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

sckott:
Thanks for raising this issue. I agree that `insert_cassette()` and `use_cassette()` should inherit from whatever the user sets with `vcr_configure()` - but that the 2 former functions can override those defaults if the user sets them. i guess we set all those parameters to default of `NULL` in `insert_cassette()` and `use_cassette()` ?

PR #153, opened 2020-02-12
Cassette config inheritance

aaronwolen:
As discussed in #151. This changes the defaults for arguments in `use_cassette()`/`insert_cassette()` to `NULL` *if* they control parameters that are also configurable via `VCRConfig`. Before making this change I reformatted the code so there's only one argument per line to make the [diff](https://github.com/ropensci/vcr/pull/153/commits/3a24c0aa3fa50ab84ccd0f739ba70c346ef71084) more readable.

Other changes: 
* Adds test helper `vcr_test_configuration()` to reset the current configuration and restore settings that should be in place for all tests. This could be called within a `teardown()` block at the top of every test file that includes modifications to the configuration. Currently it's only used in the new inheritance tests.
* Adds tests to verify that cassettes inherit settings from `vcr_configuration()`

### DIFF
--- a/R/configuration.R
+++ b/R/configuration.R
@@ -1,34 +1,17 @@
-#' Configuration
+#' Global Configuration Options
 #'
-#' Optional settings to customize vcr's default behavior.
+#' Configurable options that define vcr's default behavior.
 #'
 #' @param ... configuration settings used to override defaults. See below for a
 #'   complete list of valid arguments.
 #'
 #' @section Configurable settings:
 #'
-#' ## Casstte options
+#' ## vcr options
+#'
+#' ### File locations
 #'
 #' - `dir` Cassette directory
-#' - `record` (character) One of 'all', 'none', 'new_episodes', or 'once'.
-#' See [recording]
-#' - `match_requests_on` vector of matchers. Default: (`method`, `uri`)
-#' See [request-matching] for details.
-#' - `cassettes` (list) don't use
-#' - `linked_context` (logical) linked context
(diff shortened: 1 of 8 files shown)


sckott:
All looks good, except:

We should update the parameter docs for `insert_cassette` and `use_cassette` for those that are changed in this PR to say that their default values are set by `vcr_configure` - which only needs to be done in `use_cassette` i guess since `insert_cassette` inherits from `use_cassette`

sckott:
And document - make it explicit that any parameters set in `insert_cassette`/`use_cassette` override those set with `vcr_configure`

aaronwolen:
Cool, I'll work on updating the docs.

aaronwolen:
I took a stab at documenting the new inheritance behavior in 
6d033a7 but I feel like it could be clearer. That was the best my brain could come up with late on a Friday afternoon 🙂. Let me know if you have any suggestions. Otherwise I'll make another pass on Monday.

sckott:
looks good to me - can you resolve the conflict, then i can merge

aaronwolen:
Thanks for looking it over. I made a few additional non-coding changes that include some minor rewording/reorgs in the docs.

The most significant is cb23594, which  reorganizes the configuration docs a bit to create a separate section for cassette options that can be configured both globally and locally—and added a brief description that attempts to explain how they relate to each other. 

Also just rebased to sync up with master.

sckott:
Both look good, thanks.

Issue #152, opened 2020-02-12
helper fxn to get webmockr/vcr versions used to record each fixture


sckott:
In going through the process of submitting a new version of a package that uses vcr, i was re-recording cassettes with the newest vcr and webmockr versions on cran. And i thought it would be easier if I could tell what fixtures had been recorded with old versions of webmockr/vcr

a draft function

```r
stract <- function(str, pattern) regmatches(str, regexpr(pattern, str)) # fxn already in vcr

recorded_with <- function(dir = ".") {
  fixs <- list.files("tests/fixtures", full.names = TRUE)
  test_files <- list.files("tests/testthat/", "test-", full.names = TRUE)
  tests <- lapply(test_files, readLines)
  cas_names <- stats::setNames(lapply(tests, function(g) {
    tmp <- g[grep("use_cassette", g)]
    if (length(tmp) == 0) return(NULL)
    unlist(lapply(tmp, function(w) stract(strsplit(w, "use_cassette")[[1]][2], "[A-Za-z0-9_-]+")))
  }), test_files)

  ymls <- lapply(fixs, yaml::yaml.load_file)
  res <- stats::setNames(lapply(ymls, function(w) {
    tmp=unique(vapply(w$http_interactions, "[[", "", "recorded_with"))
  }), fixs)
  uniq_vers <- unique(unlist(unname(res)))
  
  for (i in seq_along(uniq_vers)) {
    fls <- names(res[res %in% uniq_vers[i]])
    fls_cas_nms <- gsub("tests/fixtures/|\\.yml", "", fls)
    fix_tests <- stats::setNames(lapply(fls_cas_nms, function(m) {
      chk <- vapply(cas_names, function(z) m %in% z, logical(1))
      if (any(chk)) chk[chk]
    }), fls)
    cli::cat_line(uniq_vers[i], col = "blue")
    for (i in seq_along(fls)) {
      cli::cat_line(paste0("  ", fls[i], collapse = "\n  "),
        col = "green")
      cli::cat_line(paste("    ", cli::symbol$arrow_right, names(fix_tests[[i]]), collapse = "\n  "),
        col = "red")
    }
  }
}
```

```r
recorded_with()
```

gives (with some lines removed for brevity)

```r
vcr/0.4.0, webmockr/0.5.0
  tests/fixtures/count_facet.yml
     → tests/testthat//test-count_facet.r
  tests/fixtures/dataset_metrics_error.yml
     → tests/testthat//test-dataset_metrics.r
[...]

PR #154, opened 2020-02-13
working on making errors on cassettes fails less verbose

### LINKED ISSUE #121: more brief cassette error message?
from #120 could just report that top part about the cassette and request, then say go to the book and/or  issue tracker

sckott:
- added a `vcr_configure()` option `verbose_errors` to toggle more or less verbose errors
- still needs some tweaking because some error cases dont make much sense w/o more details
    - record mode "none": uses the message in [`deal_with_none`](https://github.com/ropensci/vcr/blob/master/R/error_suggestions.R#L15-L20) - but in the default less verbose mode introduced here, that message doesn't get to the user - so some re-working needs be done
    - are there other cases where the brief/less verbose error message doesn't make sense? 
- added little function `vcr_last_error()` to get full details of the last error only - probably not that useful in practice though since i imagine most errors happen in non-interactive usage

## Related Issue

fixes #121 

## Example

example of the new error

```r
================================================================================
An HTTP request has been made that vcr does not know how to handle:
GET https://www.itis.gov/ITISWebService/jsonservice/getITISTerms?srchKey=wolf
vcr is currently using the following cassette:
  - ../fixtures/itis_terms.yml
    - record_mode: once
    - match_requests_on: method, uri
If you're not sure what to do, open an issue at https://github.com/ropensci/vcr/issues
or take a look at https://books.ropensci.org/http-testing
================================================================================
```

And you could toggle the `verbose_errors` setting to get what we have in the current version of vcr:

```r
vcr_configure(verbose_errors = TRUE)
```

```r
================================================================================
An HTTP request has been made that vcr does not know how to handle:
GET https://www.itis.gov/ITISWebService/jsonservice/getITISTerms?srchKey=wolf
vcr is currently using the following cassette:
  - ../fixtures/itis_terms.yml
    - record_mode: once
    - match_requests_on: method, uri


Under the current configuration vcr can not find a suitable HTTP interaction
to replay and is prevented from recording new requests. There are a few ways
you can deal with this:

  * If you're surprised vcr is raising this error
    and want insight about how vcr attempted to handle the request,
    you can use 'logging' to see more details [1].

  * You can use the :new_episodes record mode to allow vcr to
    record this new request to the existing cassette [2].

  * If you want vcr to ignore this request (and others like it), you can
    set an `ignore_request` function [3].

  * The current record mode ('once') does not allow new requests to be recorded
    to a previously recorded cassette. You can delete the cassette file and re-run
    your tests to allow the cassette to be recorded with this request [4].

  * The cassette contains 3 HTTP interactions that have not been
    played back. If your request is non-deterministic, you may need to
    change your 'match_requests_on' cassette option to be more lenient
    or use a custom request matcher to allow it to match [5].

[1] https://books.ropensci.org/http-testing/logging

[2] https://books.ropensci.org/http-testing/record-modes.html#new_episodes

[3] https://books.ropensci.org/http-testing/vcr-configuration#config-ignore-requests

[4] https://books.ropensci.org/http-testing/record-modes.html#once

[5] https://books.ropensci.org/http-testing/request-matching

================================================================================
```


R/configuration.R
```diff
@@ -69,6 +69,12 @@
 #' to the real data. Before record replacement happens in internal
 #' function `write_interactions()`, while before playback replacement
 #' happens in internal function `YAML$deserialize_path()`
+#' 
+#' ## Errors
+#' 
+#' - `verbose_errors` Do you want more verbose errors or less verbose
+#' errors when cassette recording/usage fails? Default is `FALSE`, that is,
+#' less verbose errors.
```
aaronwolen: Maybe add something about *suggestions* since those won't be printed either, like

* If `TRUE` error messages will include more details about what went wrong and suggest possible solutions (default `FALSE`)

R/configuration.R
```diff
@@ -270,6 +277,10 @@ VCRConfig <- R6::R6Class(
     write_disk_path = function(value) {
       if (missing(value)) return(private$.write_disk_path)
       private$.write_disk_path <- value
+    },
+    verbose_errors = function(value) {
+      if (missing(value)) return(private$.verbose_errors)
+      private$.verbose_errors <- value
```
aaronwolen: Maybe add an `is.logical()` assertion here.

R/errors.R
```diff
@@ -179,21 +183,33 @@ UnhandledHTTPRequestError <- R6::R6Class(
...
     cassettes_list = function() {
-      lines <- c()
-      xx <- if (length(cassettes_session()) == 1) {
-        "vcr is currently using the following cassette:"
+      if (length(cassettes_session()) > 0) {
+        lines <- c()
+        xx <- if (length(cassettes_session()) == 1) {
+          "vcr is currently using the following cassette:"
+        } else {
+          "vcr are currently using the following cassettes:"
```
aaronwolen: Typo, should be: vcr is

aaronwolen:
BTW, this was my first time looking over `UnhandledHTTPRequestError`, I have a few minor suggestions about its implementation that are not related to this PR at all but I'll leave them here:

* `UnhandledHTTPRequestError$new()`'s `cassette` argument is never actually used (it's only instantiated [here](https://github.com/ropensci/vcr/blob/e52e1beb7fc73e713b9d428f101b7d362beff589/R/request_handler.R#L191-L194)). Do you think we'll use it in the future? If not, we can simplify some of the logic in the `initialize()` and `run()` methods that check the argument.
* In #156 you used the cli package for colored output. If you're not opposed to adding that as a dependency, 
* `UnhandledHTTPRequestError$formatted_suggestions()` could leverage `pluck()` to simplify its implementation a bit:
    ```
    sugs <- self$suggestions()
    msg <- c(
      mapply(self$format_bullet_point, pluck(sugs, "text"), seq_along(sugs)),
      mapply(self$format_foot_note,    pluck(sugs, "url"),  seq_along(sugs))
     )
     paste0(msg, collapse = "\n", sep = "\n")
     ```

R/configuration.R
sckott: yes, will do

R/configuration.R
sckott: will do

R/errors.R
sckott: thanks, will fix

sckott:
- `cassette` argument: are you suggesting removing it?
- i assume you meant #152 - we do already use cli and crayon in suggests, that's why i used it in that issue
- that sounds like a good simplification

aaronwolen:
Ah, sorry. Ignore the cli comment. I meant to take that out and file a separate issue.

aaronwolen:
> cassette argument: are you suggesting removing it?

Yes, *if* it's not something you were planning to use in the future.

sckott:
moved those comments about `UnhandledHTTPRequestError` to a new issue

PR #156, opened 2020-02-19
Apply assertions to VCRConfig's initial arguments

aaronwolen:
This is mostly an academic change with zero functional impact for users but I think it's the "correct" approach for adding assertions to R6 classes, so it might be useful as a reference if nothing else. 

In the current implementation assertions are only applied when updating the fields of an existing object, so it's possible to initialize an object with an invalid value:

``` r
config <- vcr:::VCRConfig$new(record = "once")
config$record <- 1
#> Error in check_record_mode(value): is.character(x) is not TRUE

vcr:::VCRConfig$new(record = 1)
#> <vcr configuration>
#>   Cassette Dir: .
#>   Record: 1
#>   URI Parser: crul::url_parse
#>   Match Requests on: method, uri
#>   Preserve Bytes?: FALSE
#>   Logging?: FALSE
#>   ignored hosts: 
#>   ignore localhost?: FALSE
#>   Write disk path:
```

With this new change, initial arguments are passed to the active binding rather than being assigned directly to their private field, so assertions are applied in both cases:

``` r
config <- vcr:::VCRConfig$new(record = "once")
config$record <- 1
#> Error in check_record_mode(value): is.character(x) is not TRUE

vcr:::VCRConfig$new(record = 1)
#> Error in check_record_mode(value): is.character(x) is not TRUE
```

<sup>Created on 2020-02-19 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

You have way more experience with R6 so I'm curious to see what you think and whether I'm missing something.

### DIFF
--- a/R/configuration.R
+++ b/R/configuration.R
@@ -310,27 +310,27 @@ VCRConfig <- R6::R6Class(
       filter_sensitive_data = NULL,
       write_disk_path = NULL
     ) {
-      private$.dir <- dir
-      private$.record <- record
-      private$.match_requests_on <- match_requests_on
-      private$.allow_unused_http_interactions <- allow_unused_http_interactions
-      private$.serialize_with <- serialize_with
-      private$.persist_with <- persist_with
-      private$.ignore_hosts <- ignore_hosts
-      private$.ignore_localhost <- ignore_localhost
-      private$.ignore_request <- ignore_request
-      private$.uri_parser <- uri_parser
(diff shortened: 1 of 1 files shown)


sckott:
good catch on making sure user inputs are checked on initialize as well. 

using active binding on init seems smart to me. 

ran tests and such and don't see any issues.

PR #173, opened 2020-06-14
Cassette (class): Refactor Webmock Creation

### LINKED ISSUE #98: simplify webmock creation in `cassette_class`
way too complex https://github.com/ropensci/vcr/blob/master/R/cassette_class.R#L212-L227 and getting more complex as i'm adding body support

alex-gable:
## Description
Replaced the long `ifelse` with a set of functions. It's definitely more concise, hopefully a bit more clear, but the "why" is still opaque and could use additional commenting. 

Tests not needed unless `stub_request_with` is broken out of the initialize function. Given the lack of intuition around the rules, it might make sense

input/output used to construct the `stub_request_with` function:

|  Type      | Name           |            |         |               |             |          |            |            |               |             |            |               |             |               |               |
|------------|----------------|------------|---------|---------------|-------------|----------|------------|------------|---------------|-------------|------------|---------------|-------------|---------------|---------------|
|  Variable  | Method         |  TRUE      |         |               |             |          |  TRUE      |  TRUE      |  TRUE         |  TRUE       |  TRUE      |  TRUE         |  TRUE       |  TRUE         |  TRUE         |
|  Variable  | Uri            |            |  TRUE   |               |             |          |  TRUE      |            |  TRUE         |  TRUE       |  TRUE      |  TRUE         |  TRUE       |  TRUE         |  TRUE         |
|  Variable  | Query          |            |         |  TRUE         |             |          |            |            |  TRUE         |             |            |  TRUE         |             |  TRUE         |  TRUE         |
|  Variable  | Headers        |            |         |               |  TRUE       |          |            |            |               |  TRUE       |            |  TRUE         |  TRUE       |               |  TRUE         |
|  Variable  | Body           |            |         |               |             |  TRUE    |            |  TRUE      |               |             |  TRUE      |               |  TRUE       |  TRUE         |  TRUE         |
|  Parameter | method         |  r$method  |  any    |  any          |  any        |  any     |  r$method  |  r$method  |  r$method     |  r$method   |  r$method  |  r$method     |  r$method   |  r$method     |  r$method     |
|  Parameter | uri            |            |  r$uri  |               |             |          |  r$uri     |            |  r$uri        |  r$uri      |  r$uri     |  r$uri        |  r$uri      |  r$uri        |  r$uri        |
|  Parameter | uri_regex      |  .         |         |  .            |  .          |  .+      |            |  .         |               |             |            |               |             |               |               |
|  Case      |  use wi_th?    |            |         |  TRUE         |  TRUE       |  TRUE    |            |  TRUE      |  TRUE         |  TRUE       |  TRUE      |  TRUE         |  TRUE       |  TRUE         |  TRUE         |
|  Parameter | list$body      |            |         |               |             |  r$body  |            |  r$body    |               |             |  r$body    |               |  r$body     |  r$body       |  r$body       |
|  Parameter | list$headers   |            |         |               |  r$headers  |          |            |            |               |  r$headers  |            |  r$headers    |  r$headers  |               |  r$headers    |
|  Parameter | list$query     |            |         |  u$parameter  |             |          |            |            |  u$parameter  |             |            |  u$parameter  |             |  u$parameter  |  u$parameter  |

## Related Issue
fix #98

### DIFF
--- a/R/cassette_class.R
+++ b/R/cassette_class.R
@@ -179,6 +179,7 @@ Cassette <- R6::R6Class(
       #### first, get previously recorded interactions into `http_interactions_` var
       self$http_interactions()
       # then do the rest
+      
       prev <- self$previously_recorded_interactions()
       if (length(prev) > 0) {
         invisible(lapply(prev, function(z) {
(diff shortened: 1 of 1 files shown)


sckott:
Thanks! LGTM 

> the "why" is still opaque and could use additional commenting.

what do you mean exactly? I assume you mean the webmockr bits are not clear?

alex-gable:
re: webmockr bits unclear

Exactly. FWIW, I’m not well-versed in webmockr, so it might be more interpretable to someone who is. 

Specifically confused about the `regex_uri` patterns and why they’re dependent on the request body. For instance, in the case where body is an empty string, should we fall back to “.”?

Finally, given the complexity of the current rules and initialize method as well as the desire to add more match parameters in the future, I’m wondering if this should be broken out into its own method.

sckott:
Opened an issue to better explain what's going on wrt webmockr https://github.com/ropensci/vcr/issues/176

> Specifically confused about the regex_uri patterns and why they’re dependent on the request body. For instance, in the case where body is an empty string, should we fall back to “.”?

if the user wants to match **ONLY** on body, then we need to match on any URL

Yes, I think it could be broken out into a separate function; do you have interest/time?

PR #186, opened 2020-10-29
Stress the importance of library(vcr) in setup files and README

### LINKED ISSUE #185: using library(vcr) is actually mandatory
Hi @sckott, 

I have used `vcr` successfully today. There is one thing that was not clear to me until I looked at the code: `library(vcr)` is actually mandatory otherwise yaml files written are empty. My understanding is that is because of the initialization, i.e. the call to `.onLoad()`. 

I think it would be worth mentioning this is the documentation and in the `herper*` file created by `vcr::use_vcr()` 

```R
# a comment here?
library("vcr")
invisible(vcr::vcr_configure(
  dir = "../fixtures"
))
vcr::check_cassette_names()
```

Maybe a comment line saying that `library("vcr")` is required even though the package namespace is used below. 

Anyway, thank you for your work on this package.

KevCaz:
- fix #185 ;
- second commit of this PR fixes a few typos I found while looking at the code.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -7,7 +7,7 @@
 ### Code contributions
 
 * Fork this repo to your Github account
-* Clone your version on your account down to your machine from your account, e.g,. `git clone https://github.com/<yourgithubusername>/vcr.git`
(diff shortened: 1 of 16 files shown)


sckott:
thanks, looks good!

PR #193, opened 2020-11-17
better debugging experience?

### LINKED ISSUE #192: Improve debugging experience
From https://github.com/ropensci-books/http-testing/issues/51#issuecomment-728163070

* Should vcr switch back to creating a helper file instead of a setup file? Helper files in tests/testthat are no longer recommended but they are still supported and have the advantage of being loaded by `load_all()`. See https://deploy-preview-134--admiring-allen-9a00b2.netlify.app/2020/11/18/testthat-utility-belt/ Otherwise in docs about debugging one would simply recommend to source the setup file, which isn't bad either.

* Could vcr have a function à la testthat::test_path() that'd return the path to tests/fixtures when run in tests _and_ when run from the root of the package? This way in the configuration there would be

```
library("vcr")
invisible(vcr::vcr_configure(
  dir = vcr::test_path("fixtures") # one could still choose another location
))
vcr::check_cassette_names()


```

where `vcr::test_path("fixtures")` is `../fixtures` in tests, and `tests/fixtures` otherwise.

* Could there be a vignette about debugging explaining the above and also mentioning `insert_cassette()`? Because if you are debugging the code in a test you might prefer using `insert_cassette()` instead of `use_cassette()`, as it allows more interactive debugging (you can run the code in the test, add a line, see retrying happening whereas testthat hides messages, etc)

maelle:
Fix #192


maelle:
I am not sure how exactly to deal with the fact I used the testthat source code to create the new function.

vignettes/debugging.Rmd
```diff
@@ -0,0 +1,68 @@
...
+```
+
+If you want to run the code as if you were in the test,
+
+```r
+foo <- function() crul::ok('https://httpbin.org/get')
+vcr::insert_cassette("testing") # it will be created if needed
+x <- foo()
+x
+# further interactive debugging and fixes
```
maelle, line 60: maybe there should be more details here. E.g. would one use `http_interactions()`, logging...

sckott:
> I am not sure how exactly to deal with the fact I used the testthat source code to create the new function.

its MIT licensed, so probably fine what you've already done

R/vcr_test_path.R
```diff
@@ -0,0 +1,31 @@
...
+#'
+#' @return A character vector giving the path
+#' @export
+#'
+#' @examples
+#' @examplesIf interactive()
+#' vcr_test_path("fixtures")
+# Adapted from https://github.com/r-lib/testthat/blob/45a9c705402bd51af29b9d999e587ba789f6203f/R/test-path.R#L1
+vcr_test_path <- function(...) {
+  if (identical(Sys.getenv("TESTTHAT"), "true") && !isTRUE(getOption("testthat_interactive"))) {
```
sckott, line 16: 80 line width please, thx

R/vcr_test_path.R
```diff
@@ -0,0 +1,31 @@
+#' Locate file in tests directory
+#'
+#' This function, similar to `testthat::test_path()`, is designed to work both
+#' interactively and during tests, locating files in the `tests/` directory.
+#'
+#' @param ...	Character vectors giving path component.
```
sckott, line 6: maybe do a bit better in explaining this parameter than testthat does - i.e,. each character string gets added on to the path, e.g., `vcr_test_path("a", "b")` becomes `tests/a/b`

sckott (changes requested):
- It seems that `vcr_test_path` allows empty strings. e.g., `vcr_test_path('', 'b')` gives `tests//b`. Should that be allowed?

vignettes/debugging.Rmd
sckott: Yes, logging. We could have a separate vignette for logging or just have it in here

R/vcr_test_path.R
maelle: oops, fixed.

R/vcr_test_path.R
maelle: added

maelle:
That's the behavior of `testthat::test_path()` but I've changed it now for `vcr_test_path()`.

vignettes/debugging.Rmd
maelle: If here, would you mind adding the section? I have never used logging.

sckott (approved):

vignettes/debugging.Rmd
sckott: added

PR #195, opened 2020-11-17
provide a custom testthat skipper

### LINKED ISSUE #191: provide a custom testthat skipper?
that'd be `skip_if_vcr_off()`

maelle:
Fix #191


tests/testthat/test-skip-vcr-off.R
```diff
@@ -0,0 +1,9 @@
+test_that("skip_if_vcr_off works when vcr on", {
+  expect_silent(skip_if_vcr_off())
+})
+
+test_that("skip_if_vcr_off works when vcr off", {
+  withr::local_envvar("VCR_TURN_OFF" = TRUE)
+  expect_error(skip_if_vcr_off(), class = "skip")
```
maelle, line 7: from https://testthat.r-lib.org/articles/skipping.html#helpers

R/skip-vcr-off.R
```diff
@@ -0,0 +1,27 @@
...
+#'  you made it a 502 status code for testing the behavior of your code when the
+#'   API errors)
+#'  or if the tests are very specific (e.g. testing
+#'  a date was correctly parsed, but making a real request would produce a different
+#'  date).
+#'
+#' @return Nothing, skip test.
+#' @export
+#'
+#' @seealso lightswitch
```
sckott, line 17: link to the manual file `[lightswitch]`

R/skip-vcr-off.R
```diff
@@ -0,0 +1,27 @@
+#' Skip tests if vcr is off
```
sckott, line 1: i'd prefer to keep lines (docs and code) 80 characters or narrower - referring to all lines here

R/skip-vcr-off.R
```diff
@@ -0,0 +1,27 @@
...
+#'
+#' @seealso lightswitch
+skip_if_vcr_off <- function() {
+
+  if (is(try(find.package("testthat"), silent = TRUE), "try-error")) {
+    stop("This function is meant to be use within testthat tests. Install testthat.")
+  }
+
+  if (nzchar(Sys.getenv("VCR_TURN_OFF")) && as.logical(Sys.getenv("VCR_TURN_OFF"))) {
+    testthat::skip("Not run when vcr is off")
```
sckott, line 25: if `testthat` is used within the package, and not just for tests, does `testthat` need to be in Imports?

R/skip-vcr-off.R
```diff
@@ -0,0 +1,27 @@
...
+#'  a date was correctly parsed, but making a real request would produce a different
+#'  date).
+#'
+#' @return Nothing, skip test.
+#' @export
+#'
+#' @seealso lightswitch
+skip_if_vcr_off <- function() {
+
+  if (is(try(find.package("testthat"), silent = TRUE), "try-error")) {
```
sckott, line 20: can we use `inherits` instead of `is`?

sckott, line 20: What are relative merits of `requireNamespace` vs. `find.package`? I've never used or heard of `find.package`, so just curious

tests/testthat/test-skip-vcr-off.R
sckott, line 7: So this is intended behavior when running tests for vcr, correct?

```r
── Skip (test-skip-vcr-off.R:7:3): skip_if_vcr_off works when vcr off ─
Reason: Not run when vcr is off
```

sckott (changes requested):

R/skip-vcr-off.R
maelle: I had to add `[turn_off()]` because there was no link produced with `[lightswitch]`.

R/skip-vcr-off.R
maelle: I do not think so since I check its installation and it is only needed for testthat users (you would not want to use this function if you did not already use testthat)

R/skip-vcr-off.R
maelle: no you are right requireNamespace is better

tests/testthat/test-skip-vcr-off.R
maelle: ah maybe not https://github.com/r-lib/testthat/issues/1240

maelle:
Any further comment?

sckott (approved):

sckott:
looks good

PR #201, opened 2020-11-23
Cassette editing

### LINKED ISSUE #197: Add an article or two for covering cassette management
* When to delete a cassette and re-record it. (maybe in the debugging article? an out-of-date cassette is a common reason for test failure after all)

* how and why to edit your cassettes by hand or with a script with the `yaml` package
   * to test for error behavior without having to learn webmockr
   * to test for sequence of requests (e.g. 502 then 200) without having to learn webmockr
   * if you edit a cassette you might want to add `skip_if_vcr_off()`.
   * we imagine you will edit cassette by hand but you could also imagine writing a script to make your edits reproducible (using the yaml package if your cassettes are YAML; using jsonlite if your cassettes are JSON). However it might be more work than it's worth; editing cassettes by hand is perfectly fine.

maelle:
Fix #197

### DIFF
--- a/vignettes/cassette-manual-editing.Rmd
+++ b/vignettes/cassette-manual-editing.Rmd
@@ -131,7 +131,139 @@ testhat("Errors are handled well", {
 A big pro of this approach is that it works even when vcr is turned off.
 A con is that it's quite different from the vcr syntax.
 
-## Example 1: test using an edited cassette with a 503 then a 200
+## Example 2: test using an edited cassette with a 503 then a 200
+
+Here we assume your package contains some sort of [retry](https://blog.r-hub.io/2020/04/07/retry-wheel/).
+
+First, write your test e.g.
+
+```r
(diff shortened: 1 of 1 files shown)


vignettes/cassette-manual-editing.Rmd
```diff
@@ -0,0 +1,273 @@
...
+
+```
+
+The pro of this approach is the elegance of the stubbing, with the two different responses.
+Each webmockr function like `to_return()` even has an argument `times` indicating the 
+number of times the given response should be returned.
+
+The con is that on top of being different from vcr, in this case where we also needed 
+a good response in the end (the one with a 200 code, and an actual body), writing the
+mock is much more cumbersome than just recording a vcr cassette.
```
maelle, line 266: this sentence is the reason why I am so excited about editing vcr cassettes :grin:

sckott (approved):
looks good!

PR #202, opened 2020-11-26
better vignette index in pkgdown?

maelle:
* All vignette titles now start with a capital letter and are a sentence
* Add an order from easy/most basic usage to advanced.

### DIFF
--- a/vignettes/configuration.Rmd
+++ b/vignettes/configuration.Rmd
@@ -1,5 +1,5 @@
 ---
-title: "Tweak vcr default configuration"
+title: "Configure vcr"
 author: "Scott Chamberlain"
 date: "`r Sys.Date()`"
 output:
--- a/vignettes/configuration.Rmd
+++ b/vignettes/configuration.Rmd
@@ -1,5 +1,5 @@
 ---
-title: "vcr configuration"
+title: "Tweak vcr default configuration"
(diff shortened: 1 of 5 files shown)


vignettes/configuration.Rmd
```diff
@@ -1,5 +1,5 @@
 ---
-title: "vcr configuration"
+title: "Tweak vcr default configuration"
```
sckott, line 2: Could we use a different word than tweak? would "manage vcr configuration" be better?

vignettes/request_matching.Rmd
```diff
@@ -1,5 +1,5 @@
 ---
-title: "request matching"
+title: "Tweak vcr request matching"
```
sckott, line 2: similar comment, could we maybe not use tweak

sckott (changes requested):
In the vignette title "Why and how edit your vcr cassettes?" could we change to "Why and how __to__ edit your vcr cassettes?"

maelle:
I made changes. Now I wonder whether it'd be clearer if request matching were merged into "configure vcr" as it's part of configuration?

sckott:
No, I think request matching just deserves more attention than most other configuration settings - we should at some point fill out that matching vignette more

sckott (approved):

PR #204, opened 2020-11-27
Encoding YAML Windows

### LINKED ISSUE #203: is there an alternative to preserve_exact_body_bytes for Windows?
Asking because the cassettes are much less nice to look at in that case.

Editing them by hand would not be fun.

maelle:
Fix #203?

In https://github.com/maelle/vcrencoding/actions/runs/386871714 I am using this branch and things work on Windows

This PR

* adds a test that fails on Windows with vcr current version (see this commit https://github.com/ropensci/vcr/pull/204/commits/d315686b8ae9105e9559d30954603236330c52a1)
* adds a fix for this test

I wasn't able to find a more elegant way to add the test.


R/serializers-yaml.R
```diff
@@ -37,9 +37,14 @@ YAML <- R6::R6Class("YAML",
...
-      str <- sensitive_put_back(readLines(self$path))
-      tmp <- yaml::yaml.load(str)
+      tmp <- yaml_load_desecret(self$path)
       private$process_body(tmp)
     }
   )
 )
+
+yaml_load_desecret <- function(path) {
+  str <- sensitive_put_back(readLines(path, encoding = "UTF-8"))
```
maelle, line 47: `encoding = "UTF-8"` here is key.

tests/testthat/cassettes/ropenaq-encoding.yaml
```diff
@@ -0,0 +1,869 @@
+http_interactions:
```
maelle, line 1: a bit big for sure but I don't know which part provoked the failure.

sckott:
thanks, i'll have a look

tests/testthat/cassettes/ropenaq-encoding.yaml
sckott: that's okay

PR #209, opened 2020-12-08
more vignettes, less duplication

### LINKED ISSUE #4: Add test suite

maelle:
This PR

* adds vignettes that are chapters in the HTTP testing book (then, the chapters from the HTTP testing book will read content from this repo)
* uses the lightswitch Rmd fragment in both the lightswitch vignette and lightswitch man page (less duplication).

### DIFF
--- a/vignettes/write-to-disk.Rmd
+++ b/vignettes/write-to-disk.Rmd
@@ -27,7 +27,6 @@ Request matching
 
 ```{r setup}
 library("vcr")
-vcr::turn_on()
 ```
 
 ```{r child='../man/rmdhunks/write-to-disk.Rmd', eval=TRUE} 
--- a/vignettes/write-to-disk.Rmd
+++ b/vignettes/write-to-disk.Rmd
@@ -27,6 +27,7 @@ Request matching
 
 ```{r setup}
 library("vcr")
+vcr::turn_on()
 ```
 
 ```{r child='../man/rmdhunks/write-to-disk.Rmd', eval=TRUE} 
--- a/vignettes/write-to-disk.Rmd
+++ b/vignettes/write-to-disk.Rmd
@@ -8,6 +8,7 @@ output:
     toc_float: true
     theme: readable
 vignette: >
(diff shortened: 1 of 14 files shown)


vignettes/write-to-disk.Rmd
```diff
@@ -0,0 +1,39 @@
...
+    message = FALSE
+)
+```
+
+Request matching
+================
+
+```{r setup}
+library("vcr")
+vcr::turn_on()
```
maelle, line 30: Actually it's not very good to have this line here. I think it is because the lightswitch vignette turns off vcr. Could the lightswitch vignette "clean after itself"?

sckott: What do you suggest?

sckott (commented):
looks good, just that one question

vignettes/write-to-disk.Rmd
maelle: not sure, maybe adding a vcr::turn_on at the end of the lightswitch vignette. What confuses me is: aren't vignette knit in a clean environment?

sckott: Not sure about the clean env thing.  Not sure what exactly is wrong here. The vignettes build just fine right?

maelle: Without the vcr::turn_on() above there was an error saying vcr was not enabled.

sckott: Okay, probably related to previous vignette settings. Can we just put it in a different block and not show it (echo=FALSE or whatever it is)

sckott: any thoughts @maelle ?

maelle: I'll experiment in a minimal package next week, I want to understand more about how a vignette can influence another vignette.

maelle: yes, there's no new content after all, only doc infrastructure change.

sckott (approved):

PR #215, opened 2020-12-14
Update configuration-vignette.Rmd

maelle:
less confusing I think?

i.e. with the current example you'd write an environment variable in cassettes.


sckott:
yes, thanks

sckott (approved):

PR #216, opened 2020-12-17
make cassette names content re-usable

maelle:
What's not optimal is that includeRmd does not work for user-defined sections so I had to remove the section title. https://roxygen2.r-lib.org/articles/rd.html#sections-1

I then want to use this fragment in the book for https://github.com/ropensci-books/http-testing/issues/17

### DIFF
--- a/man/use_cassette.Rd
+++ b/man/use_cassette.Rd
@@ -134,28 +134,6 @@ block.
 \code{use_cassette()} block starts on.
 }
 
-\section{Cassette names}{
-
-\itemize{
-\item Should be meaningful so that it's obvious to you what test/function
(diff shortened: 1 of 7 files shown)


man/use_cassette.Rd
```diff
@@ -134,28 +134,6 @@ block.
 \code{use_cassette()} block starts on.
 }
 
-\section{Cassette names}{
```
sckott, line 137: is it intended that use_cassette loses the __Cassette names__ section?

sckott (commented):
one comment

man/use_cassette.Rd
maelle: See my first comment, I can't avoid this

PR #229, opened 2021-03-16
Update vcr_test_path.R

### LINKED ISSUE #228: Does vcr cache things somewhere?
Weird behavior seen by @dpprdan and me

* Install vcr dev version
* Use https://github.com/maelle/exemplighratia/pull/2, run tests, then delete fixtures and turn off wifi, run tests: they still pass!

maelle:
Fix #228

### DIFF
--- a/R/vcr_test_path.R
+++ b/R/vcr_test_path.R
@@ -24,7 +24,7 @@ vcr_test_path <- function(...) {
     if (missing(...)) {
       "../."
     } else {
-      here::here("..", ...)
+       file.path("..", ...)
     }
   } else {
     path <- here::here("tests", ...)


maelle:
@dpprdan mentioned it might create bugs according to your last commit on the file?

sckott:
what is "it"? which commit?

dpprdan:
In https://github.com/ropensci/vcr/commit/2e07d00d8502e8c78b2a595c880de76094f97fd7 you wrote: 

> when we used file.path we got relative paths and those would fail in some scenarios

dpprdan:
I wonder, whether this would suffice:

```r
vcr_test_path <- function(...) {
  if (any(!nzchar(...))) {
    stop("Please use non empty path elements.")
  }
  root <- rprojroot::is_r_package
  path <- root$find_file("tests", ...)
  if (!dir.exists(path)){
    message("could not find ", path, "; creating it")
    dir.create(path)
  }
  path
}
```

In the current function, 

- the `missing` condition is never reached (because `nzchar(...)` does not allow missing and you probably do not want the cassettes directly in `tests` dir). 
- the `path` is not created (but `message`d).

When using `here()` everything is relative to the project root, so we don't need all the conditions (I assume). Also, we always want the cassettes to be below the `tests` dir (from the docs "locating files in the `tests/` directory").

Using {rprojroot}, because the {here} [docs state](https://here.r-lib.org/reference/here.html): 

> This package is intended for interactive use only. Use `rprojroot::has_file()` or the other functions in the rprojroot package for more control, or for package development.

The tests need to be adjusted accordingly. I could attempt to make a PR, if you wish.

sckott:
I was wondering about the `missing` path, but I figured Maëlle had a good reason for it. Makes sense to drop it. 

> the path is not created (but messaged).

I think this is because the path is created by vcr the package in general, not this function specifically. 

I like idea of using `rprojroot` instead of `here`. 

A PR would be great.

dpprdan:
This is only intended for packages, correct? I.e. [`rprojroot::is_r_package`](https://rprojroot.r-lib.org/reference/criteria.html) is sufficient?

sckott:
Well, vcr in general is not specific to pkgs, but `vcr_test_path` can be, yes

dpprdan:
👍🏻 Yes, that was my understanding.

sckott:
@dpprdan I released the change from `vcr_test_path` from this PR and https://github.com/ropensci/vcr/pull/230 in the latest vcr version on cran - it's caused problems for a number of packages.  See https://github.com/ropensci/vcr/issues/235  I don't use the function in my pkgs so I didn't see any issues. We need to get this sorted somewhat quickly

PR #230, opened 2021-03-16
simplify vcr_test_path()

dpprdan:
## Description
This simplifies `vcr_test_path()` as discussed in #229. It uses {rprojroot} instead of {here}. It also adjusts `vcr_test_path()`-related tests accordingly.

## Related Issue
#228, #229


sckott:
thanks! having a look

sckott:
Looks good to me. Tested locally and on some packages that use vcr

Issue #235, opened 2021-05-24
CRAN message about broken reverse dependency checks


jsta:
> Subject: CRAN packages archiveRetriever dbhydroR deepdep hlidacr intensegRid weathercan
Today's vcr update 1.0.0 has broken the checks of these.

All tests pass locally for me (running `devtools::test`) but fail running as part of `devtools::check`. The error I get is the same as CRAN:

> Error: No root directory found in /home/hornik/tmp/R.check/r-devel-gcc/Work/PKGS/dbhydroR.Rcheck/tests/testthat or its parent directories. Root criterion: contains a file "DESCRIPTION" with contents matching "^Package: "

I was able to pass check associated tests by removing the call to `vcr_test_path` in the autogenerated "setup" file and hardcoding the path to the fixtures directory.

jsta:
`dir = vcr::vcr_test_path("fixtures")` -> `dir = "../fixtures"`

sckott:
Thanks. Was going to send an email to people this morning. I did run revdep checks and didn't get any issues (and I don't personally use the function in my own http pkgs), but I should have emailed about this issue, sorry about that. Looking into it

dpprdan:
👋🏻 My PR #230 probably caused this. Sorry for the trouble this is causing!

I can reproduce with https://github.com/ropensci/opencage/tree/devel. `devtools::check()`/R CMD check apparently moves the test dir outside the pkg dir: 

```
.
+-- 00check.log
+-- 00install.out
+-- 00_pkg_src
|   \-- opencage
+-- opencage
|   +-- data
|   +-- DESCRIPTION
|   +-- doc
|   +-- extdata
|   +-- help
|   +-- html
|   +-- INDEX
|   +-- Meta
|   +-- NAMESPACE
|   +-- NEWS.md
|   \-- R
+-- opencage-Ex.pdf
+-- opencage-Ex.R
+-- opencage-Ex.Rout
+-- opencage-Ex.timings
\-- tests
    +-- startup.Rs
    +-- testthat
    +-- testthat.R
    +-- testthat.Rout.fail
    \-- vcr_cassettes
```

(`vcr_cassettes` is what is usually called `fixtures`)

I looking into ways how to catch this. Maybe `rprojroot::has_dir("tests")`?

dpprdan:
[...]

Issue #225, opened 2021-02-15
vcr writes empy cassettes if not run using Run Tests


llrs:
When I run (just Ctrl+return) a single `test_that` that makes use of `vcr::use_cassette` I get an empty cassette, while when I run it using the Rstudio Run Tests then they are recorded and the cassette has all the information of the requests and reply (Rstudio Version 1.4.1103).

Perhaps this is related to the [note](https://docs.ropensci.org/vcr/index.html#vcr-for-tests): "some users have found different results with `devtools::check()` vs. `devtools::test()`. It’s not clear why this would make a difference"

If #224 is fixed this could be prevented.

<details> <summary><strong>Session Info</strong></summary>

```r
? Session info ????????????????????????????????????????????????????????????????????????
 setting  value                       
 version  R version 4.0.1 (2020-06-06)
 os       Ubuntu 20.04.2 LTS          
 system   x86_64, linux-gnu           
 ui       RStudio                     
 language (EN)                        
 collate  en_US.UTF-8                 
 ctype    en_US.UTF-8                 
 tz       Europe/Madrid               
 date     2021-02-15                  

? Packages ????????????????????????????????????????????????????????????????????????????
 ! package     * version   date       lib source        
   assertthat    0.2.1     2019-03-21 [1] CRAN (R 4.0.1)
   base64enc     0.1-3     2015-07-28 [1] CRAN (R 4.0.1)
   BiocManager * 1.30.10   2019-11-16 [1] CRAN (R 4.0.1)
 R bugRzilla   * 0.0.90001 <NA>       [?] <NA>          
   cachem        1.0.3     2021-02-04 [1] CRAN (R 4.0.1)
   callr         3.5.1     2020-10-13 [1] CRAN (R 4.0.1)
   cli           2.3.0     2021-01-31 [1] CRAN (R 4.0.1)
   crayon        1.4.1     2021-02-08 [1] CRAN (R 4.0.1)
   crul          1.0.0     2020-07-30 [1] CRAN (R 4.0.1)
   curl          4.3       2019-12-02 [1] CRAN (R 4.0.1)
   cyclocomp   * 1.1.0     2016-09-10 [1] CRAN (R 4.0.1)
   desc          1.2.0     2018-05-01 [1] CRAN (R 4.0.1)
   devtools    * 2.3.2     2020-09-18 [1] CRAN (R 4.0.1)
   ellipsis      0.3.1     2020-05-15 [1] CRAN (R 4.0.1)
[...]

Issue #162, opened 2020-03-04
check if vcr works with tinytest


sckott:

sckott:
started on this on branch https://github.com/ropensci/vcr/compare/tinytest

maelle:
did it work?

sckott:
don't remember, so long ago now. i don't think it worked, OR I didn't fully test it yet, not sure which one

maelle:
Ok

dpprdan:
tinytest stores/expects `test_*.R` files in [`inst/tinytest`](https://github.com/markvanderloo/tinytest/tree/master/pkg#package-setup), rather than `tests/tinytest`, so at the very least we'd have to adapt `vcr_test_path()` to that (assuming that tinytest also sets the working dir to `inst/tinytest` during tests).

jimrothstein:
@sckott 
Be happy to work on tinytest::  testing.    Also have worked with httr:: and httr2:: 
Will fork clone and  use the branch already setup https://github.com/ropensci/vcr/compare/tinytest

jim

sckott:
thanks @jimrothstein ! sounds great

jimrothstein:
@sckott 
Ok, on my end deleted fork and started again.
When I pushed, got referenced here "c648c0a", which should be correct code.  But I don't see way to create new PR.

jimrothstein:
@sckott 
working just locally, writing  **`tinytest** functions  roughly equivalent to **`testthat`** functions  ( so far for `test_Filesystem.R` and `test_Cassette.R`)   When I have couple more files  tested will try again to push.

sckott:
great!

jimrothstein:
@sckott   Quick update:   tinytests are coming along, but reading through some of the source code, R6, configuration takes some time.

sckott:
ok, thanks for the update

jimrothstein:
@sckott 
Update.    This is rough, rough, very rough.  WIP.
But tinytests are beginning to work and various pieces make more sense to me.
Will get faster.
Example:
https://github.com/jimrothstein/test_vcr/blob/master/inst/tinytest/test_configure.R

sckott:
@jimrothstein  sounds good

jimrothstein:
@sckott 
Bunch of tinytests seems to be working.    Before I get too much further, I like to get my package {testVCR} into form you can review.   (There are still loose ends or temp files that need cleanup.   And, of course, many more tests to write.)

[...]

PR #236, opened 2021-05-24
vcr_test_path() fix: look for 'tests' dir instead of DESCRIPTION

dpprdan:
- DESCRIPTION is not available for tests run by R CMD check, #235

Both `devtools::test()` and `devtools::check()` seem to run successfully with this change. 

Caveat: If `vcr_test_path()` is run from a file `setup-pkg.R` in `./tests/testthat/` (i.e. the recommended setup, AFAIU),  `has_dir()` is starting to search for the `tests` dir from `./tests/testthat/` and go further up the dir tree. As a result, it will use `./tests/testthat/tests/` or `./tests/tests/` as a "root" dir should those exist and the test path will start from there.

### DIFF
--- a/R/vcr_test_path.R
+++ b/R/vcr_test_path.R
@@ -16,7 +16,7 @@
 vcr_test_path <- function(...) {
   if (missing(...)) stop("Please provide a directory name.")
   if (any(!nzchar(...))) stop("Please use non empty path elements.")
-  root <- rprojroot::is_r_package
+  root <- rprojroot::has_dir("tests")
   path <- root$find_file("tests", ...)
   if (!dir.exists(path)){
     message("could not find ", path, "; creating it")


sckott:
Thanks! 

I'll ping some of the authors of the other affected pkgs and see if any of them can test this out.

dpprdan:
@sckott Is this sufficient, given aforementioned caveat, or should we consider alternatives?

Possible alternatives:

1. `rprojroot::is_testthat` ("Root criterion: directory name is "testthat" (also look in subdirectories: `tests/testthat`, `testthat`)"): Similar caveat as `rprojroot::has_dir("tests")`, works only with {testthat}. (Doesn't find the package root, but `testthat` dir within `./tests`)
2. `rprojroot::has_file("testthat.R")` ("Root criterion: contains a file "testthat.R"): similar to 1, probably lower risk to match on a dir further down the path (assuming that users might use multiple dirs called tests more often than using multiple `testthat.R` files). 
3. Go back to using relative paths with `file.path()`, possibly `normalizePath()`ing them. We'd have to find a better solution for #225, then. 

Looking at the function description, `has_dir("tests")` probably comes closes to what `vcr_test_path()` is intended to do:

> This function, similar to `testthat::test_path()`, is designed to work both interactively and during tests, locating files in the `tests/` directory.

So I could just add a warning to the docs, 'This might not work if you have more than one directory named "tests" in your package'?

Somewhat relatedly: Should vcr be compatible with test suites other than testthat, e.g. tinytest? IIUC tinytest's test files are located in `/inst/tinytest/`, so the cassettes probably should be, too?

sckott:
I guess considering we want `vcr` to be useable with other testing pkgs, we probably don't want 1 or 2, and I don't think we want 3 given the #225 issue 

The solution in this PR works for me in all scenarios I've tested.

> As a result, it will use ./tests/testthat/tests/ or ./tests/tests/ as a "root" dir 

I don't see this when I tried it locally. I went from the root of a pkg and ran the two lines https://github.com/ropensci/vcr/pull/236/files#diff-3f0a5d7d447e7c5c839094838571d9e1f38b5ecfa19e057193ba32c5959c6851R19-R20, then went into `tests/` and ran them again, and also within `tests/testthat` - and those two lines all gave back the correct path (e.g., when run for my zbank pkg within `tests/testthat/` I get `"/Users/sckott/github/ropensci/zbank/tests/fixtures"`) 

> So I could just add a warning to the docs, 'This might not work if you have more than one directory named "tests" in your package'?

Seems unlikely, right? So we probably don't have to worry about that situation very much 

---  

We do have an issue already for tinytest/testit https://github.com/ropensci/vcr/issues/162 - I don't think we want to address this now - it should be thoroughly tested, and we don't want to rush it i think

dpprdan:
> I don't see this when I tried it locally.

Sorry, I did not describe this clearly.

> it will use `./tests/testthat/tests/` or `./tests/tests/` as a "root" dir _should those exist_

Emphasis added now. So if there is a second `tests` dir within the `tests` dir (the one in the package root) it will use the one further down the dir tree (though depending on the current working directory, I suppose). 
E.g. if I have a `tests` dir within `/tests/testthat` and I run `devtools::test()` with the vcr setup in `/tests/testthat/setup-mypackage.R´ I'll end up with

```
.
+-- [...]
+-- tests
|   +-- testthat
|   |   +-- setup-mypackage.R
|   |   +-- [...]
|   |   +-- tests
|   |   |   \-- fixtures
|   |   |       +-- some_vcr_cassette.yml
|   |   |       +-- [...]
|   \-- testthat.R
\-- vignettes
    +-- [...]

```

instead of

```
.
+-- [...]
+-- tests
|   +-- testthat
|   |   +-- setup-opencage.R
|   |   +-- [...]
|   |   \-- tests
|   |-- testthat.R
|   \-- fixtures
|       +-- some_vcr_cassette.yml
|       +-- [...]
\-- vignettes
    +-- [...]
```

Then again, how likely is it that users will put `tests` within `tests`? 

> it should be thoroughly tested, and we don't want to rush it i think

Definitely! My idea behind the tinytest question was that with tinytest the cassettes shouldn't live in `/tests` but in `/inst/tests` (because that's where the tinytest test files are AFAIK). So the `has_dir("tests")` solution does not work in that case. But we can think about that, when we tackle tinytest.

sckott:
@dpprdan  Thanks for the explanation. I'll see if I can have a quick look to answering this tests within tests question

sckott:
```r
library(crul)
library(glue)

con <- HttpClient$new("https://api.github.com",
  headers = list(Authorization = paste0("token ", Sys.getenv("GITHUB_PAT"))))


# vcr reverse dependencies
revdeps <- c("bold", "chirps", "circle", "citecorp", "crminer", "dbhydroR", "deepdep", "finbif", "fulltext", "gfonts", "hlidacr", "intensegRid", "jaod", "magmaR", "microdemic", "natserv", "pangaear", "qualtRics", "rbace", "rbhl", "rbison", "rcol", "rcoreoa", "rcrossref", "rdatacite", "rdryad", "rerddap", "rgbif", "ritis", "rnoaa", "rnpn", "rorcid", "rphylopic", "rplos", "rredlist", "rromeo", "rstac", "rvertnet", "Rwtss", "spocc", "taxize", "weathercan", "webmockr", "wikitaxa", "worrms", "zbank")

# get latest tree sha
tree_sha <- function(pkg, org = "cran", branch = "master") {
  z <- con$get(glue("repos/{org}/{pkg}/branches/{branch}"))
  res <- jsonlite::fromJSON(z$parse("UTF-8"))
  res$commit$commit$tree$sha
}

# get tree
tree <- function(sha, repo, org = "cran") {  
  path <- glue("repos/{org}/{repo}/git/trees/{sha}")
  w <- con$get(path, query = list(recursive = "true"))
  temp <- jsonlite::fromJSON(w$parse("UTF-8"))
  temp$tree$path
}

# any test(s) dir within tests?
any_nested_tests <- function(paths) {  
  paths_in_test <- grep("^tests/", paths, value = TRUE)
  any(grepl("^tests/testthat/tests?/", paths_in_test))
}

# run all pkgs
res <- list()
for (i in seq_along(revdeps)) {
  ts <- tree_sha(revdeps[i])
  tr <- tree(ts, revdeps[i])
  res[[ revdeps[i] ]] <- any_nested_tests(ts)
}
any(unlist(res))
#> FALSE
```

And I tested it with a repo I created https://github.com/sckott/testintest to make sure a path `tests/testthat/tests/` is indeed found

So at least none of the current revdeps have `tests/testthat/tests/` - could run over all of cran pkgs to see if any pkgs do this

sckott:
@dpprdan I think we just need to add to the vcr_test_path docs about the tests within tests thing and then we're good to go

dpprdan:
@sckott Sorry I did not follow up! I was busy with other stuff and overlooked this. Feel free to ping me next time! Great that this is solved for now.

sckott:
Thanks for helping get this fixed

PR #243, opened 2021-12-18
set vcr_test_path() relative to testthat dir

### LINKED ISSUE #242: vcr_test_path() might need an update
Cc @adamhsparks @dpprdan 

It seems tests sometimes are run in

```
:/RCompile/CRANguest/R-release/packagename.Rcheck/tests_i386/testthat
```

So `tests_i386` not `tests`.

dpprdan:
`vcr_test_path()` now looks for the `testthat` dir instead of `tests`, because the latter may not be present during `R CMD check`. 
`fixtures` (or however the folder is called by the user) is then put on the same level as `testthat` (instead of one level up from `tests`). 

fixes #242 (hopefully)

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -52,4 +52,4 @@ Suggests:
 X-schema.org-applicationCategory: Web
 X-schema.org-keywords: http, https, API, web-services, curl, mock, mocking, http-mocking, testing, testing-tools, tdd
 X-schema.org-isPartOf: https://ropensci.org
-RoxygenNote: 7.1.1
+RoxygenNote: 7.1.2
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,10 @@
+vcr 1.0.3
+=========
+
+### BUG FIXES
+
+* fix to `vcr_test_path()` to find root package path correctly with R 4.2 on Windows (#242) (#243)
+
 vcr 1.0.2
 =========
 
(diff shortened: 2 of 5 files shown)


maelle:
This sounds like a great idea!
Should the docs state this will only work with testthat? I'm not sure there are vcr users who don't use testthat, though.

dpprdan:
> Should the docs state this will only work with testthat?

I'd say the docs are pretty clear that vcr ~~works only with~~ assumes tests are done with testthat, see e.g. the first sentence in the [Getting Started section](https://docs.ropensci.org/vcr/#getting-started). OTOH it doesn't hurt to mention it explicitly.

dpprdan:
I just noticed that the README/Getting Started vignette still use

```r
invisible(vcr::vcr_configure(
  dir = "../fixtures"
))
```
i.e. `"../fixtures"` to specify the fixtures directory instead of `dir = vcr::vcr_test_path("fixtures")` [as mentioned](https://books.ropensci.org/http-testing/vcr.html) in the http-testing book.

Should we change that?

dpprdan:
BTW, if I am right about #244, or rather that #225 wasn't caused by `vcr_test_path()` - and that's a big IF - then we might be able to use a `testthat::test_path()` based approach after all and drop the rprojroot dependency. (IIUC we only began fiddling with `vcr_test_path()` because of #225).

This is based on multiple hunches though, so I'd go with the `rprojroot::find_testthat_root_file()` for now. We will have to revisit this topic again anyway, when we want to adapt vcr for other test suites (#162).

sckott:
looks good to me. I don't use `vcr_test_path` so if others could please make sure to test this in their packages.

Issue #249, opened 2022-11-06
Error thrown for cassettes that have an empty body


KevCaz:
Hi @sckott,

It looks like there is something wrong with cassettes that have a status code that is not 200. I only tried 401 and 404 so not sure about the other codes. (EDIT: @dpprdan pointed out that this is true for cassettes that have an empty body) Here is a reprex:

```r
usethis::create_package("testpkg")
usethis::use_package("httr")
vcr::use_vcr()
file.remove("tests/testthat/test-vcr_example.R")

writeLines("
test_that('default', {
  vcr::use_cassette('get_401', {
    res <- httr::GET('https://httpbin.org/status/401')
  })  
  expect_true(res$status_code == 401)
})

", con = "tests/testthat/test.R")
```
 
The first time `devtools::test()` is used, it works fine, the cassette is recorded correctly. But the second time I get this:

```r
R> devtools::test()
ℹ Testing testpkg
✔ | F W S  OK | Context
✖ | 1       0 | test                                                                                                                                                  
─────────────────────────────────────────────────────────────────
Error (test.R:3): default
Error in `vapply(interactions, function(x) {
    sprintf("%s => %s", request_summary(Request$new()$from_hash(x$request)), 
        response_summary(VcrResponse$new()$from_hash(x$response)))
}, "")`: values must be length 1,
 but FUN(X[[1]]) result is length 0
Backtrace:
 1. vcr::use_cassette(...)
      at test.R:3:2
 2. vcr::insert_cassette(...)
 3. Cassette$new(...)
 4. vcr (local) initialize(...)
 5. self$http_interactions()
 6. HTTPInteractionList$new(...)
 7. vcr (local) initialize(...)
 8. base::vapply(...)
```

I did not investigate further because I have limited time to work on this right now, I thought I could post this now in case you have an idea how to solve it quickly but if you cannot work on this now I could also investigate latter (assuming there is actually something wrong). 

<details> <summary><strong>Session Info</strong></summary>

```r
sessionInfo()
R version 4.2.2 (2022-10-31)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 22.04.1 LTS

Matrix products: default
BLAS:   /usr/lib/x86_64-linux-gnu/openblas-pthread/libblas.so.3
LAPACK: /usr/lib/x86_64-linux-gnu/openblas-pthread/libopenblasp-r0.3.20.so

locale:
[...]

PR #252, opened 2022-11-07
Use again body_from for the body of the response

### LINKED ISSUE #4: Add test suite

llrs:
## Description

This PR reverts some changes as discussed in #249 to handle the body again with the `body_from` function and adds a new test for it.

## Related Issue

#249 

## Example

I'm not sure how to add tests for this, but I added the test as reported in #249 in a new file test-issues.

Tests and checks pass but I'm not sure if the issue is fully resolved.

### DIFF
--- a/R/response_class.R
+++ b/R/response_class.R
@@ -42,7 +42,7 @@
 #' x$headers$`content-length` # = NULL
 #' x$update_content_length_header() # no change, b/c header doesn't exist
 #' x$headers$`content-length` # = NULL
-#' 
+#'
 #' ## example 3
 #' ### content-length header present, and does change
 #' body <- " Hello World "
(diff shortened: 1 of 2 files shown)


sckott:
@llrs can you add back `body_from` in request class as well. In the offending commit https://github.com/ropensci/vcr/commit/cf5ee5bc88b1bb840e8b7c0d6e1a556151b0454c#diff-a947054c47f80b7252dac92d05ce9294c02a7671022a290403913f0bcfa4c197R150-R151 it was removed from both request and response class

sckott:
that test I don't think really tests the change - not sure quite how to do it yet, but maybe we merge and think about that later

llrs:
Before the change it didn't pass and after the change it did, but I wasn't sure if it was the best way to test the effects (I knew it was not minimal).

KevCaz:
@sckott @llrs  maybe this is rather minimal and reflect the changes: 

```R
body <- NULL
x <- VcrResponse$new(200, list('content-length' = nchar(body)), body, "HTTP/2") # tookk this piece from one of your example

test_that('issue 249 is correctly handled.', {
  expect_equal(x$from_hash(x$to_hash())$body, character(1))
})
```

before it should have been NULL now it's "".

sckott:
thanks @KevCaz - may add that after

PR #254, opened 2022-11-10
Tinytest::  setup to begin adding tests #162

jimrothstein:
## Description
Added tinytest::
-  added to DESCRIPTION
-  added simple test, test_addOne.R  in inst/tinytest/
-  added tests/tinytest.R so package is aware of tinytest::

## Related Issue
issue #162, just 1st step

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -48,7 +48,8 @@ Suggests:
     crayon,
     cli,
     curl,
-    withr
+    withr,
+    tinytest
 X-schema.org-applicationCategory: Web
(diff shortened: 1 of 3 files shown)


sckott (commented):
Thanks @jimrothstein 

- I don't know what the recommended file placement is for tinytest. What is it? I ask because I imagine a test file shouldn't be in the root of a package? Or should it?
- I think perhaps we want to take a different approach: make a small package that's super simple, and that uses `tinytest` and `vcr` for its tests, then we can go from there to see what changes need to be made to `vcr`.

PR #262, opened 2023-06-24
Update cassette-editing-vignette.Rmd

adamhsparks:
Fixes code: replaces `testhat()` with `test_that()`
Corrects grammar: replaces "a first" with "the first"

No new tests since this is just a vignette.


sckott:
thank you @adamhsparks  !

PR #280, opened 2025-04-08
Switch to testthat 3e

hadley:
And switch parallel testing on, radically improving the time needed to run all tests (I also dropped the wait time in `use_cassette_re_record` from 10s to 3s).


sckott:
Nice, thank you, Yeah much faster tests.

PR #283, opened 2025-04-09
Mildly polish insert/eject API

hadley:
* Return objects invisibly
* Use early returns in `inject_cassette()` to make flow more clear


sckott:
Thanks! Tests pass locally.

PR #284, opened 2025-04-09
Slightly simplify `current_cassette()`

hadley:
My eliminating not very helpful `last()` function. Also make tests a bit more comprehensive.


sckott:
Thanks. Need to do more on this one - there's a fair amount of behavior that depends on no cassette in use being an empty list `list()` ...

hadley:
You can (a) use `pr_fetch()` to get this branch locally and then push back to it or (b) tell me to the fix the problems 😄

PR #285, opened 2025-04-09
Implement `local_vcr_configure()` helper

hadley:
This required changing `vcr_configure()` to make it work like `options()`, namely if called with no options return a list of all options, and if called with some new options, return the previous values of those options.

I used it a few test just to get a sense for how it works.


sckott:
thanks! made a small fix

PR #286, opened 2025-04-09
Cache httpbin server in an option

hadley:
The helpers are called on `load_all()`, so the previous approach occasionally caused package loading to take a long time. Saving it in an option means that we only ever need to compute it once per session. In the long run, it would be better to switch to `webfakes::httpbin_app()`, but this will require some work because webfakes doesn't implement exactly the interface that vcr expects.


sckott:
Thanks! 

I'll open an issue for `webfakes::httpbin_app()`

PR #289, opened 2025-04-11
Only run tests once in `R CMD check`

hadley:
Looks like I accidentally added this in #280, because `use_testthat()` didn't recognise the older file structure.


sckott:
Thanks for catching that, I missed that at the time

PR #290, opened 2025-04-11
Consistently namespace functions

hadley:
This had the unintended (but possibly useful/good) side effect that httr and htt2 are no longer hard dependencies, which does make sense to me.

2eb4851a8bc6dae8ba0239d2526ea0dd461971b8 was made by ✨ claude code ✨ for a cost of $2.69.


sckott:
Thanks! 

Hmm, shouldn't we use `skip_if_not_installed` for `httr` and `httr2` since they're in Suggests now (when either of those pkgs is not installed).

hadley:
@sckott don't those functions only get hit if you make a httr or httr2 request?

hadley:
FWIW the tidyverse is not to use `skip_if_not_installed()`, unless the package is particularly hard to install. In general, you can expect that any one testing the package is developer and has all the hard and soft requirements installed.

sckott:
@hadley 

> don't those functions only get hit if you make a httr or httr2 request?

yes, so we're all set I think

Okay, good to know best practice on `skip_if_not_installed()`.

sckott:
@hadley  I'm happy to reimburse you for the claude costs. Let me know.

hadley:
@sckott no worries! This is useful for me learning more about it.

sckott:
Okay thanks.

PR #293, opened 2025-04-11
Use base64 encoding and decoding from jsonlite

hadley:
Dropping the base64env dependency. jsonlite was already a hard dependency via crul, httr, and webmockr.


sckott:
thanks!

PR #298, opened 2025-04-11
redact request headers automatically for httr2

### LINKED ISSUE #277: Automatically scrub redacted headers from cassettes
This is only possible with `httr2` as there are not mechanisms like `httr2:: req_headers_redacted` in `httr` or `crul`

There are various filters in `vcr` to do this, but could document that for this case it will automatically be done

Replacement could just be `<redacted>`

sckott:
fix #277

### DIFF
--- a/R/filter_headers.R
+++ b/R/filter_headers.R
@@ -55,6 +55,5 @@ request_headers_redact <- function(x) {
 request_headers_redact_one <- function(headers) {
   redacted_headers <- attr(headers, "redact")
   headers[intersect(redacted_headers, names(headers))] <- "<redacted>"
-  }
   headers
 }
--- a/R/filter_headers.R
+++ b/R/filter_headers.R
@@ -54,11 +54,7 @@ request_headers_redact <- function(x) {
 
 request_headers_redact_one <- function(headers) {
   redacted_headers <- attr(headers, "redact")
-  if (is_empty(redacted_headers)) return(headers)
-  for (i in seq_along(headers)) {
-    if (names(headers)[i] %in% redacted_headers) {
-      headers[[i]] <- "<redacted>"
-    }
+  headers[intersect(redacted_headers, names(headers))] <- "<redacted>"
   }
   headers
 }
--- a/R/filter_headers.R
+++ b/R/filter_headers.R
@@ -44,3 +44,21 @@ headers_remove <- function(x) {
   x <- filter_req_or_res(x, vcr_c$filter_request_headers, "request")
   filter_req_or_res(x, vcr_c$filter_response_headers, "response")
 }
+
+request_headers_redact <- function(x) {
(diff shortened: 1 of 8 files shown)


sckott:
@hadley try this branch. Note that this only impacts what's written to the cassette, not the objects returned in the R session. 

A different approach could be to hook into the filter system https://docs.ropensci.org/vcr/articles/configuration.html#filter_request_headers but I couldn't see a way for that to work on a shallow investigation so went with the approach here

I still need to do:

- [ ] add mention of this to docs in all relevant places
- [ ] add test for json serialization to make sure that works. don't expect any issues

R/filter_headers.R
```diff
@@ -44,3 +44,21 @@ headers_remove <- function(x) {
...
+  })
+}
+
+request_headers_redact_one <- function(headers) {
+  redacted_headers <- attr(headers, "redact")
+  if (is_empty(redacted_headers)) return(headers)
+  for (i in seq_along(headers)) {
+    if (names(headers)[i] %in% redacted_headers) {
+      headers[[i]] <- "<redacted>"
+    }
```
hadley, line 61:
```suggestion
  headers[intersect(redacted_headers, names(headers))] <- "<redacted>"
```
?

tests/testthat/test-redact-httr2-request-headers.R
```diff
@@ -0,0 +1,62 @@
...
+
+test_that("redact request headers works when needed", {
+  use_cassette("redacted_correctly", {
+    httr2::request(hb("/get")) %>%
+      httr2::req_headers_redacted(
+        SecretHeader = "Hidden",
+        AnotherSecret = "SecretKey"
+      ) %>%
+      httr2::req_perform()
+  })
```
hadley, line 25: IMO you could combine this with the previous test by doing something like:

```R
httr2::request(hb("/get")) %>%
  httr2::req_headers(NotASecret = "one") %>%
  httr2::req_headers_redacted(SecretHeader = "Hidden"))
```

hadley (approved):
Nice, thanks!

tests/testthat/test-redact-httr2-request-headers.R
```diff
@@ -0,0 +1,45 @@
+local_vcr_configure(dir = withr::local_tempdir())
```
hadley, line 1: One last point — I think there's some advantage to doing this inside of each individual test, as it keeps the tests completely independent and means that you get the same results when you're iterating on an individual test. That's the direction I'm slowly heading in with my other test refactors.

sckott: Thanks, I'd missed that. Will do!

sckott:
found an issue with json serialization, fixing that

PR #301, opened 2025-04-12
More work making tests self-contained

hadley:
(I think I've got a couple more PRs in this vein, and then I'll add a `testthat::set_state_inspector()` so we can verify that the tests actually are self-contained)


sckott:
Looks good. 

> then I'll add a `testthat::set_state_inspector()` so we can verify that the tests actually are self-contained

Sounds good, thanks

PR #302, opened 2025-04-13
More work making tests self-contained

hadley:
Introduced new helper to ensure that changes to `light_switch` state are kept local to the test.


sckott:
thanks! looks good.

PR #303, opened 2025-04-13
Use local httpbin test server where possible

hadley:
This gives us the foundation to eventually eliminate the reliance on external services as we either update the vcr code and/or add functionality to webfakes.


sckott:
thanks!

PR #304, opened 2025-04-13
Silence webmockr disable

hadley:
Since this message comes from webmockr, it's not obvious what relevance it has to the vcr user and I think it's clearer if you omit it.


sckott:
Makes sense

PR #305, opened 2025-04-14
Use `format()` instead of `as.character()`

hadley:
This ensures that the timestamp always has the same number of characters, nicely aligning the log lines. Otherwise (e.g.) `as.character(as.POSIXct("2002-1-1"))` gives `2002-01-01`.


sckott:
thanks, makes sense

PR #306, opened 2025-04-14
Implement `local_cassette()`

hadley:
And use it to implement `use_casette()`. Since this eliminates most of the reasons to use `insert_cassette()` and `eject_cassette()` I combined theirs docs, and made them internal.

I haven't updated the recommend test practices yet because I want to make a few more changes first.


R/cassette_class.R
```diff
@@ -14,12 +14,6 @@
 #' - `eject()` method: [webmockr::disable()] is called before exiting
 #' eject to disable webmock so that webmockr does not affect any HTTP
 #' requests that happen afterwards
-#' - `call_block()` method: call_block is used in the [use_cassette()]
-#' function to evaluate whatever code is passed to it; within call_block
-#' [webmockr::webmockr_allow_net_connect()] is run before we evaluate
```
hadley, line 19: This doesn't seem to be true any more, anyway.

R/use_cassette.R
```diff
@@ -175,17 +179,45 @@ use_cassette <- function(
...
-  cassette$call_block(...)
-  return(cassette)
+
+  # force all arguments
+  list(...)
+
+  invisible(cassette)
 }
 
-check_empty_cassette <- function(cas) {
```
hadley, line 187: I deleted this function since it's not used anywhere. (The equivalent is now performed by `Casette$remove_empty_cassette()`)

R/zzz.R
```diff
@@ -60,6 +60,8 @@ can_charToRaw <- function(x) {
 
 stp <- function(...) stop(..., call. = FALSE)
 check_cassette_name <- function(x) {
+  if (length(x) != 1) stp("cassette name must be a single string")
```
hadley, line 63: Not really related but I noticed that this was missing, so `use_cassette(letters)` generated an uninformative error.

sckott: thanks!

sckott:
> Since this eliminates most of the reasons to use insert_cassette() and eject_cassette() I combined theirs docs, and made them internal.

So should we remove those two functions from pkgdown yml?

R/eject_cassette.R
```diff
@@ -9,8 +7,8 @@
 #' `allow_unused_http_interactions = FALSE` cassette option. This is intended
 #' for use when your test has had an error, but your test framework has
 #' already handled it - IGNORED FOR NOW
-#' @return The ejected cassette if there was one
-#' @seealso [use_cassette()], [insert_cassette()]
```
sckott, line 13: there's another seealso at line 95

R/use_cassette.R
```diff
@@ -175,17 +179,45 @@ use_cassette <- function(
...
-  if (!any(nzchar(readLines(cas$file())))) {
-    warning(empty_cassette_message, call. = FALSE)
+#' @rdname use_cassette
+#' @export
+#' @param frame For expert use only: the frame on the call stack that defines
```
sckott, line 191: Can we have an example for how to use frame or maybe point to docs for https://withr.r-lib.org/reference/defer.html ?

sckott:
thanks

PR #310, opened 2025-04-14
Maintain a stack of cassettes

hadley:
This PR makes two major changes so that active cassettes form a stack:

* The data structure (now `the$cassettes`) that records which cassettes are active is now a list rather than environment.
* `eject_cassette()` no longer has a `cassette` argument as it's only allowed to remove the most recently added cassette.

I've also moved all the state management out of the `Cassette` class and into `insert_cassette()` and `eject_casette()`. I think this makes it easier to understand the code because changes to global state now only happen in functions specifically to change that state.


hadley:
Note that the `Cassette()` no longer automatically registers itself as an active cassette — registering and unregistering is now exclusively the job of `insert_cassette()` and `eject_cassette()`. I think this makes it easier to understand what's going on by isolating changes to global state to functions that are called specifically to change that state.

sckott:
@hadley can you resolve this? I don't want to resolve this more complicated one in a way you hadn't intended

hadley:
@sckott yeah, of course!

sckott:
merging some of the other PRs while I take some time to look at this one

sckott:
@hadley  Playing with this and a questions arose

- Should `eject_cassette()` say which cassette was ejected? For example, when you have more than one cassette in use (as per the tests https://github.com/ropensci/vcr/pull/310/files#diff-6a49a0444f8f04937ac7c42c3cbb9601ac434d35ec1cc6b92bb0b06c04054aa4R16) the user shouldn't have to remember the logic we use, but rather we tell them what was ejected. Thoughts?

hadley:
Maaaaaybe. My main reservation is that sounds awfully like the messaging that `summarise()` did, where it told you which groups remained, and people really did not like that. My sense is also that few enough people use nested cassettes that it's probably not worth worrying about too much.

hadley:
Also IMO 99.9% of all cassette management should be done by `local_cassette()` and `use_cassette()` so there should be little confusion about what cassettes are in play.

sckott:
Yeah ok lets leave it as is. Best to solve for the majority use case

PR #314, opened 2025-04-15
Polishing the lightswitch

hadley:
The most important change is that the env vars are only read once on load. I think this is better than re-reading (and overwriting any values set by the user) each time you call `insert_cassette()`

Since I was in here I polished the docs and the tests, refreshed the error messages, and hopefully made the env var logic a bit easier to follow.


sckott:
thanks, no notes. 

thanks for the better error messages.

PR #317, opened 2025-04-16
Drop Serializers registry

hadley:
In favour of a simpler implementation of `serializer_fetch()`. I assume the goal here was to allow users to implement their own serializer class, but that never happened. I think it's ok for vcr to not make this extensible outside of the package in favour of making the code inside the package simpler.


sckott:
👍🏽

PR #318, opened 2025-04-16
Drop Persisters

hadley:
There's only one currently supported, and I don't think the existing tooling is sufficiently generalised that code add a new one without a bunch of work.


sckott:
Correct. Good to remove

PR #319, opened 2025-04-16
Drop `update_content_length_header`

hadley:
In general, there's no expectation that the content-length should match the length of the body, because there might have been compression at the HTTP transport level.


sckott:
thanks, makes sense to remove

PR #321, opened 2025-04-16
Store the full path in the Serializer

hadley:
This makes the Serializer responsible for the entire location of the output, keeping the path and name + ext in the same place.

I also removed the docs for Serializer and its subclasses since these are for internal use only, and I think they're fairly straightforward if you pick more descriptive names for the arguments.

Finally, I made a `dir_create()` helper that sets the defaults that you always want and used it everywhere.


sckott:
looks good thanks!

PR #324, opened 2025-04-18
Logging revamp

hadley:
* Add logging specific config helper
* Create single logging function that also does `sprintf()`, and don't export it since it's not user facing.
* Don't wrap `<none>` in quotes.

@sckott let me know if this feels too aggressive and I can back it off a bit. But it makes it really easy to turn on interactive logging so you can see what's going on. I'll have more feedback for specific bits as I used it to understand how vcr and ellmer interact. (I've already discovered one problem which I'll submit a fix for shortly).


hadley:
I think it makes sense to switch to cli eventually, but I'll do that in series of PRs where I start to test the specific log output so we can iterate on exactly what it should look like.

PR #327, opened 2025-04-18
Check casettte names on initialization

hadley:
* Move check from `insert_cassette()` to `Cassette$new()`.
* Move existing cassette check into `check_cassette_name()`
* Switch to cli for error messages
* Reduce documentation — IMO folks just need to know that this is a file name, not every rule that determine if it's a valid filename or not.


sckott:
thanks, this all makes sense, and yeah users dont need to have the docs for the rules

PR #329, opened 2025-04-18
Simplify request ignorer

hadley:
It took me a little while to understand the state manipulation here, but I think there are a few layers of redundancy that can now be removed in favour of having one global `RequestIgnorer`.


hadley:
Actually I just figured out a way to simplify this even further.

sckott:
I have been wanting to support hooks concept in general like Ruby vcr does (see Hooks in sidebar on https://benoittgt.github.io/vcr/#/) 

I think the hooks stuff removed here didn't actually work and wasn't used, but I do think it would be good to support hooks as they do seem useful - but i'll open an issue for that

hadley:
Yeah, that's totally reasonable.

PR #331, opened 2025-04-18
Store `light_switch` in `the`

hadley:
And rename `turned_off` to `on` to make it a bit easier to follow the logic


sckott:
looks good thnaks

PR #333, opened 2025-04-18
Simplify RegisterMatches

hadley:
This also simplifies the `Request` class by moving more logic into the matcher. This means that that request no longer needs to store the uri components, and that port stripping is handled by the matcher.

I also removed the `uri_parser` option in favour of using `curl::curl_url_parse()` which should be both fast and correct.


hadley:
@sckott this is now ready for review

sckott:
Thanks, I think main needs to be merged here as there was a typo that makes this break https://github.com/hadley/vcr/commit/fd199e7755942fe331bf260b1ff87ab985b9c6da

R/request_matcher_registry.R
```diff
@@ -1,107 +1,55 @@
...
-      self$register("body_as_json", function(r1, r2) {
-        tc <- tryCatch(
-          identical(jsonlite::fromJSON(r1$body), jsonlite::fromJSON(r2$body)),
-          error = function(e) e
-        )
-        if (inherits(tc, "error")) FALSE else tc
-      })
-    }
+request_matches_one <- function(type, req1, req2) {
+  match <- switch(
```
sckott, line 21: Did you mean `switch(` without assigning to a var? It doesn't matter I don't think just curious

sckott (commented):
nice, much cleaner. one question

R/request_matcher_registry.R
hadley: Oops yes, this was a remnant of a previous approach. Let me remove it.

sckott: thanks

PR #334, opened 2025-04-18
Allow cassette functions to provide directory directly

hadley:


sckott:
is the idea to do away with global configuration?

hadley:
Not to get rid of it, but to allow you to work without it, if you want (which is I think what I'd want inside a test)

sckott:
> which is I think what I'd want inside a test 

yeah makes sense, i think especially if you're just sprinkling in use of vcr

PR #335, opened 2025-04-18
Drop unused utils functions

hadley:


DESCRIPTION
```diff
@@ -30,6 +30,7 @@ BugReports: https://github.com/ropensci/vcr/issues
 Imports:
     cli,
     crul (>= 0.8.4),
+    curl,
```
hadley, line 33: Already a hard dep of curl, so doesn't increase net dependencies.

R/insert_cassette.R
```diff
@@ -31,7 +31,7 @@ insert_cassette <- function(
   }
   # enable webmockr
   webmockr::enable(quiet = vcr_c$quiet)
-  sup_mssg(vcr_c$quiet, webmockr::webmockr_allow_net_connect())
+  suppressMessages(webmockr::webmockr_allow_net_connect())
```
hadley, line 34: Technically a behaviour change, but unlikely to affect many folks. I think probably we can make `vcr_c$quiet` go away in favour of better logging tools.

DESCRIPTION
sckott: 👍🏽

PR #339, opened 2025-04-19
Decode and encode body in the serializer

hadley:
* Delete `body_from()` and `serialize_body()` in favour of always encoding/decoding the body in the `Serializer`.
* Create new `encode_body()` and `decode_body()`
* Move binary tests to `test-serializer.R`

I can't swear that I've implemented exactly what you had before but I'm pretty sure I've copied the intent.


sckott:
thanks, looks good

PR #341, opened 2025-04-19
`Cassette` simplifications

hadley:


R/cassette_class.R
```diff
@@ -320,22 +317,8 @@ Cassette <- R6::R6Class(
     #' @description is the cassette on disk empty
     #' @return logical
     is_empty = function() {
-      nchar(self$raw_cassette_bytes()) < 1
-    },
-
-    #' @description timestamp the cassette was originally recorded at
-    #' @return POSIXct date
-    originally_recorded_at = function() {
-      as.POSIXct(self$recorded_at, tz = "GMT")
```
hadley, line 329: This is always saved as a `POSIXct`, so no need to coerce

R/cassette_class.R
```diff
@@ -394,13 +377,11 @@ Cassette <- R6::R6Class(
     #' @return logical
     should_re_record = function() {
       if (is.null(self$re_record_interval)) return(FALSE)
-      if (is.null(self$originally_recorded_at())) return(FALSE)
```
hadley, line 397: Could never be true

R/cassette_class.R
```diff
@@ -441,32 +422,13 @@ Cassette <- R6::R6Class(
       self$record == "all"
     },
 
-    #' @description Get character string of entire cassette; bytes is a misnomer
-    #' @return character
-    raw_cassette_bytes = function() {
```
hadley, line 446: Only used by `is_empty()` which now just looks to see if the file exists

R/cassette_class.R
```diff
@@ -441,32 +422,13 @@ Cassette <- R6::R6Class(
...
-      if (is.null(file)) return("")
-      tmp <- readLines(file) %||% ""
-      paste0(tmp, collapse = "")
-    },
-
-    #' @description get http interactions from the cassette via the serializer
-    #' @return list
-    deserialized_hash = function() {
-      tmp <- self$serializer$deserialize(self)
-      if (inherits(tmp, "list")) {
```
hadley, line 457: It's the responsibility of the deserializer to error if there's a problem, and there are lots of potential problems that we don't currently error for.

R/cassette_class.R
```diff
@@ -320,22 +317,8 @@ Cassette <- R6::R6Class(
...
-    #' @return POSIXct date
-    originally_recorded_at = function() {
-      as.POSIXct(self$recorded_at, tz = "GMT")
-    },
-
-    #' @description Get a list of the http interactions to record + recorded_with
-    #' @return list
-    serializable_hash = function() {
-      list(
-        http_interactions = self$merged_interactions(),
```
hadley, line 336: This was the only component ever actually used from this function.

R/cassette_class.R
```diff
@@ -622,54 +585,8 @@ Cassette <- R6::R6Class(
         disk = is_disk
       )
       HTTPInteraction$new(request = request, response = response)
-    },
-
-    #' @description Make a crul response object
-    #' @return a crul response
-    serialize_to_crul = function() {
```
hadley, line 629: Not used anywhere

R/cassette_class.R
sckott: yep, good move

R/cassette_class.R
sckott: i like it

PR #345, opened 2025-04-20
Don't store HTTP version

hadley:
It's not currently used, and I don't think it's going to be generally very useful for testing.


hadley:
Done!

sckott:
thanks yeah agree to remove

PR #349, opened 2025-04-20
Sanitize uri during `prep_interaction()`

hadley:
This is the first PR in a series of PRs that moves logic out of `write_interactions()` and into `prep_interactions()`. This should also make the code a little faster since eventually we'll only need to loop through the list of interactions once.

### DIFF
--- a/R/filter_query_parameters.R
+++ b/R/filter_query_parameters.R
@@ -1,33 +1,32 @@
 #' Remove or replace query parameters
 #' @noRd
-query_params_remove <- function(int) {
-  h <- vcr_c$filter_query_parameters
-  if (!is.null(h)) {
-    if (is.null(names(h))) toremove <- unlist(h)
-    if (!is.null(names(h))) toremove <- unname(unlist(h[!nzchar(names(h))]))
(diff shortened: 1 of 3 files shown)


R/filter_query_parameters.R
```diff
@@ -1,33 +1,32 @@
...
+  to_remove <- filter[!is_named]
+  for (i in seq_along(to_remove)) {
+    uri <- drop_param(uri, to_remove[[i]])
+  }
+
+  to_replace <- filter[is_named]
+  for (i in seq_along(to_replace)) {
+    val <- to_replace[[i]]
+    if (length(val) == 2) {
+      if (flip) {
```
hadley, line 19: I don't understand why this is needed, but the tests, as well as careful comparison of `query_params_remove()` and `query_params_remove_str()` revealed that it's necessary.

sckott: don't remember, I'll have a look

sckott:
@hadley  can you resolve?

hadley:
Done!

PR #350, opened 2025-04-20
Sanitize headers during `prep_interaction()`

hadley:
Again, we can save a bunch of work by relying on the iteration that's already performed by `prep_interaction()`.


sckott (changes requested):

sckott:
thanks

PR #351, opened 2025-04-21
Update eject logging

hadley:
@sckott I think this was a non-obvious merge conflict. I added a test to make sure it doesn't happen again, and also added a parallel insert log entry.


sckott:
thanks

PR #357, opened 2025-04-22
Improve file names

hadley:
* By convention, `.onLoad()` is usually found in `zzz.R`
* The existing `zzz.R` would be better called `utils.R` (and that matches the existing `test-utils.R`


sckott:
> By convention, .onLoad() is usually found in zzz.R

wasn't aware of that, sounds fine

PR #358, opened 2025-04-22
Unify cassette manipulation functions

hadley:
And get test coverage to 100%. This made me notice a few small things:

* `the$cassettte()` doesn't need to be named
* The test for `!turned_on()` should live with other `insert_cassette()` tests, and it should use `cli::cli_abort()`.
* `eject_cassette()` error was missing a trailing `.`


sckott:
makes sense to unify them in a single place

PR #365, opened 2025-04-22
Add `warn_on_empty` field to cassette

hadley:
This allows us to eliminate the use of `local_vcr_configuration()` in a number of tests. While I was touching these tests, I also took the opportunity to review where we could use other new tools like `local_cassette()`, and where we no longer need to set the cassette dir since we only write a file if an interaction is recorded.


sckott:
@hadley can you resolve?

hadley:
Done!

sckott:
looks good

PR #367, opened 2025-04-22
Move `sensitive_remove()` into encoding for body/uri/headers

hadley:
A few small changes along the way:

* Rename `query_params_remove()` to `encode_uri()`.
* Trim quotes in `vcr_configure()`.
* Rename `prep_interaction()` to `encode_interactions()`.


sckott:
thanks, all looks good

PR #369, opened 2025-04-22
Drop `to_hash()` and `from_hash()`

hadley:
It seems like these aren't currently needed because the request object and corresponding list are similar enough.


sckott:
looks good, thanks

PR #375, opened 2025-04-23
Refactor `RequestHandler`

hadley:
* Remove a layer of indirection and call private methods directly
* Use `self$request` rather than passing `request` object again


sckott:
thanks

PR #376, opened 2025-04-23
Drop `HTTPInteraction` R6 class

hadley:
Because it's now just equivalent to a list containing the request and response


sckott:
thanks, i was just going to propose we do that

hadley:
I think we'll be able to do the same with the request and response classes in a few steps too.

PR #377, opened 2025-04-23
Drop `NullList`

hadley:
* `parent_list` is not actually used
* `http_interactions()` hasn't worked properly since `current_cassette()` stopped erroring, and can be inlined in the two places it's used


sckott:
thanks 👍🏽

PR #379, opened 2025-04-23
Drop unusued `Request()` fields

hadley:
The primary job of the request object is to support request matching, so these extra fields are not needed.


sckott:
thanks

PR #382, opened 2025-04-24
Eliminate `write_` helpers

hadley:
Now we have `encode_interactions()` which creates the R data structure that can be written directly to yaml/json. I also hoisted the `recorded_by` field up a level, so it can be written once, rather than once for each interaction.


sckott:
Thanks!

PR #386, opened 2025-04-24
Re-organise files related to serialization

hadley:
* Both JSON and YAML serializers move into `serializers.R`
* A new series of files for functions encoding to and decoding from the on disk serialization:
  * `serialize-body.R`
  * `serialize-headers.R`
  * `serialize-uri.R`
  * `serialize-interaction.R`


sckott:
thanks i like the new organization

PR #390, opened 2025-04-24
Use RequestHandler request instead of creating anew

hadley:
Also inline `$make_http_interaction()` into `$record_http_interaction()` and improve argument names. I also dropped the `recursive` argument which should do anything (since it's a path, not a directory) because it makes the call fit on one line.

### DIFF
--- a/R/cassette_class.R
+++ b/R/cassette_class.R
@@ -360,9 +360,70 @@ Cassette <- R6::R6Class(
     #' @description record an http interaction (doesn't write to disk)
     #' @param x a crul, httr, or httr2 response object, with the request at `$request`
     #' @return nothing returned
-    record_http_interaction = function(x) {
-      int <- self$make_http_interaction(x)
-      self$http_interactions_$response_for(int$request)
+    record_http_interaction = function(response, request) {
+      # for httr2, duplicate `body` slot in `content`
+      if (inherits(response, "httr2_response"))
+        response$content <- response$body
+
+      # content must be raw or character
+      assert(unclass(response$content), c('raw', 'character'))
+      new_file_path <- ""
+      is_disk <- FALSE
+      if (is.character(response$content)) {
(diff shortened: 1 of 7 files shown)


sckott:
thanks, can you resolve?

R/cassette_class.R
```diff
@@ -366,100 +366,86 @@ Cassette <- R6::R6Class(
     #' @description record an http interaction (doesn't write to disk)
     #' @param x a crul, httr, or httr2 response object, with the request at `$request`
```
sckott, line 367:
```suggestion
    #' @param response a crul, httr, or httr2 response object
    #' @param request a `Request` object
```

R/cassette_class.R
```diff
@@ -366,100 +366,86 @@ Cassette <- R6::R6Class(
...
     #' @return nothing returned
```
sckott, line 368:
```suggestion
    #' @return an interaction as a list with request and response slots
```

R/cassette_class.R
```diff
@@ -366,100 +366,86 @@ Cassette <- R6::R6Class(
...
-    #' @return logical
-    any_new_recorded_interactions = function() {
-      length(self$new_recorded_interactions) != 0
-    },
-
-    #' @description Make a request-response pairs
-    #' @param x A crul, httr, or httr2 response object, with the request at `$request`
-    #' @return A list with request and response.
-    make_http_interaction = function(x) {
+    record_http_interaction = function(response, request) {
```
sckott, line 369: Wouldn't it be more consistent for the order of the param inputs to match those of the outputs? Maybe it doesn't matter too much since it's not user facing

hadley: Oh yeah, good call.

hadley:
Yup, should be better now hopefully.

sckott:
thanks!

PR #391, opened 2025-04-24
Improve Request/Response serialisation

hadley:
* Move Request/Response object creation out of `cassette_class` and into `decode_interaction()`
* Consistently record empty bodies as `NULL`
* Drop unused encoding value


sckott:
thanks, can you resolve?

PR #396, opened 2025-04-25
Implement compressed serializer

### LINKED ISSUE #187: Implement Compressed serializer

sckott:
fix #187 

This was pretty easy after the refactoring that's been happening, so thanks :)

Comments/questions

- To not have to change `sensitive_put_back` I converted the list interaction to json then character (though maybe `as.character` isn't needed), then stored in qs format - then upon reading it'd be a string so we can use `sensitive_put_back` as is. 
- I haven't yet thought about what is the most efficient in terms of storage in a `qs2` file, i.e., is it smaller file size if we store as string or list (or something else). If users wanted to they could mess with `qs2` settings https://github.com/qsbase/qs2/?tab=readme-ov-file#global-options-for-qs2 for example `use_alt_rep` which apparently improves string read time in `qs_read`
- `Compressed` is not all uppercase like the other 2 since `COMPRESSED` seems a bit loud, seem okay?
- Any other thoughts here?


R/serializer.R
```diff
@@ -75,3 +75,29 @@ YAML <- R6::R6Class(
     }
   )
 )
+
+Compressed <- R6::R6Class(
```
hadley, line 79: I'd suggest calling this qs2 to match the class/extension names for the other serializers.

R/serializer.R
```diff
@@ -75,3 +75,29 @@ YAML <- R6::R6Class(
...
+  inherit = Serializer,
+  public = list(
+    initialize = function(path, name, preserve_bytes = FALSE) {
+      check_installed("qs2")
+      super$initialize(path, name, ".qs2", preserve_bytes = preserve_bytes)
+    },
+
+    serialize = function(data) {
+      out <- encode_interactions(data, self$preserve_bytes)
+      json <- as.character(jsonlite::toJSON(out, auto_unbox = TRUE))
```
hadley, line 90: IMO it'd be better to save the R object directly. That will require changes to `decode_interactions()` to apply `sensitive_put_back()` in the opposite way that `encode_interactions()` does with `sensitive_remove()`. (That's a refactoring on my to do list)

tests/testthat/test-serializer.R
```diff
@@ -268,3 +268,75 @@ test_that("Windows encoding", {
...
+  )
+
+  request <- Request$new(method = "GET", uri = "http://example.com")
+  response <- VcrResponse$new(status = 200L, list(name = "val"), body = "body")
+  interaction <- list(request = request, response = response)
+
+  ser <- Compressed$new(withr::local_tempdir(), "serialize")
+  ser$serialize(list(interaction))
+
+  expect_snapshot(writeLines(qs2::qs_read(ser$path)))
```
hadley, line 293: I don't think I'd bother with this snapshot tests since this is fundamentally a binary file format; there's no way to easily confirm that you're getting the correct output written to disk. (But I also don't think it's necessary because the majority of the work is now done in `encode_interactions()`.

tests/testthat/test-serializer.R
```diff
@@ -268,3 +268,75 @@ test_that("Windows encoding", {
...
+})
+
+test_that("Compressed usage", {
+  skip_on_cran()
+  local_vcr_configure(
+    dir = withr::local_tempdir(),
+    serialize_with = "compressed"
+  )
+
+  # does one request work?
```
hadley, line 303: I think it would be good to think through your testing goals here, as I think this code mostly tests mechanics outside of serialization. I think a better test (which we might also want to port to YAML and JSON) would be to test that deserialisation + serialisation is idempotent.

hadley (commented):
Nice! You probably want to include a news bullet, and it might be worth noting some of the file size benefits?

R/serializer.R
sckott: using `qs2` does match the extension, but will people know what it means? I think json and yaml are widely known so that seems different. but i guess the docs will say what it does, so could be fine

tests/testthat/test-serializer.R
sckott: Good point

R/serializer.R
sckott: Okay, I did consider that, but yeah the `sensitive_put_back` issue. 

> (That's a refactoring on my to do list)

I'll leave a FIXME comment for now until that's addressed

tests/testthat/test-serializer.R
sckott: I'll have a think about that and try replacing this

R/serializer.R
hadley: Yeah, and I worry compressed might be a bit too general — what happens if qs3 comes along and is way better than qs2.

sckott: right, okay, will change to qs2

tests/testthat/test-serializer.R
sckott: @hadley thoughts on the tests now?

hadley (approved):
This looks great!

sckott:
thanks!\ 🙏🏽

PR #397, opened 2025-04-25
Improve match logging

hadley:
I've switched to using waldo here in order to generate minimal diffs in the case of a failed match. I'm not committed to this precise log format, but this is where I ended up with.

Long run, I think it would be useful to include the index of the recorded request in the log message, but that will require some `HTTPInteractionList` refactoring first.


sckott:
i'm open to using `waldo`  - it seems like a pretty lite weight dep. 

This is odd, tests are passing and all, but i'm seeing an issue: 

```r
library(vcr)
library(httr)
local_vcr_configure_log()
local_cassette("autheg789")
GET("http://example.com")
#> Error:
#> ! object 'adapter' not found

rlang::last_error()
#> <error/rlang_error>
#> Error:
#> ! object 'adapter' not found
#> ---
#> Backtrace:
#>      ▆
#>   1. └─httr::GET("http://example.com")
#>   2.   └─httr:::request_perform(req, hu$handle$handle)
#>   3.     └─httr:::perform_callback("request", req = req)
#>   4.       └─webmockr (local) callback(...)
#>   5.         └─webmockr::HttrAdapter$new()$handle_request(req)
#>   6.           └─private$request_handler(req)$handle()
#>   7.             ├─vcr:::vcr_log_sprintf(...)
#>   8.             │ └─base::sprintf(message, ...)
#>   9.             └─webmockr::enabled(adapter)
#>  10.               └─adapter %in% webmockr_adapters
```

I think with this commit https://github.com/ropensci/vcr/commit/411c6a291079f62703cf126124900b544bbf3514#diff-1416cc9e6aaf741effe014bd85d16f93dc74bd8b7a516c33bc77850b095625ccR43-R52 `adapter` is now never defined

Perhaps we add an `adapter` field to `RequestHandler` with default of `""` or `NULL` and then each of httr, httr2, and crul adapters define as the pkg name, then we refer to it in this `enabled` call and anywhere else needed

hadley:
Ooops. Given that we only enable webmockr as a whole, I think we can just remove that argument. I'll submit a PR momentarily.

sckott:
> Or maybe webmockr status doesn't need to be part of this log message at all?

I don't think it needs to be

sckott:
> the index of the recorded request in the log message,

what do you mean by the "index"?

hadley:
Like "This is request 2 in the stored cassette"

PR #398, opened 2025-04-25
Refactor `HTTPInteractionList()`.

hadley:
* Simplify the implementation by maintaining a logical vector that records when an interaction is used.
* Tweaked logging to just recorded number of interactions loaded (for now; will come back to logging holistically soon)
* Reimplemented response look up from first principles, using a single helper.

### DIFF
--- a/R/http_interaction_list.R
+++ b/R/http_interaction_list.R
@@ -10,8 +10,7 @@ HTTPInteractionList <- R6::R6Class(
     initialize = function(
       interactions,
       request_matchers = c("method", "uri"),
-      allow_playback_repeats = FALSE,
-      used_interactions = list()
+      allow_playback_repeats = FALSE
     ) {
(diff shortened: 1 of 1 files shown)


R/http_interaction_list.R
```diff
@@ -1,184 +1,65 @@
...
-    #' @param used_interactions (list) Interactions that have been used. That is,
-    #' interactions that are on disk in the current cassette, and a
-    #' request has been made that matches that interaction
-    #' @return A new `HTTPInteractionList` object
     initialize = function(
       interactions,
-      request_matchers,
+      request_matchers = c("method", "uri"),
       allow_playback_repeats = FALSE,
       used_interactions = list()
```
sckott, line 14: `used_interactions` should be removed, yes?

sckott: done

PR #403, opened 2025-04-26
Finish decoding refactoring

hadley:
* Now have `decode_sensitive()` and `encode_sensitive()` and `decode_sensitive()` is applied during `decode_interaction()`.
* New `decode_headers()`
* Rename `query_params_put_back()` to `decode_uri()` and simplify implementation

This makes serialization and deserialization more parallel and should make #396 slightly simpler to implement.


sckott (commented):
looking over these decode/encode functions I wonder if we should - while we're here - rename some of the function parameters to be more meaningful. e.g., `encode_interactions`/`decode_interactions`, `decode_sensitive`/`encode_sensitive`, and `encode_interaction`/`decode_interaction` all have first or only parameter `x` (I'm sure I'm the one that used `x`!)

hadley:
Yeah, I can do that since I'm in here.

sckott:
Thank you!

PR #414, opened 2025-04-27
Standardise request and response classes

hadley:
I've added in our new standard `check_` functions and used them to verify that the inputs to `vcr_response()` and `vcr_request()` are consistent. This led to a few changes:

* `status` is now always an integer. This simplified a lot of accessor code.
* `disk` is now either `TRUE` or `FALSE`. This also simplified a lot of accessor code.
* Multipart request bodies are always recorded as list, rather than sometimes a string and sometimes a list. This doesn't make much difference in practice but I think makes it a bit easier to understand what's going on.

While I was in here, I also tweaked the way that bodies are encoded, so that there's one field name for each of the possible body types: `on_disk`, `fields`, `base64_string` (raw), and `string`. I did this in a backward compatible way so existing cassettes will still work.


sckott:
@hadley  This is the first I've seen of these standalone files. I see https://usethis.r-lib.org/reference/use_standalone.html Is the idea that it didn't make sense for `rlang` to export these functions so it provides these standalone files? And what is the workflow for updating them? Do watch for changes in these files in the `rlang` repo? Sorry for the probably basic questions. I'm not questioning their use just trying to understand how they're used

hadley:
I don't think there's likely to be an update to these files. I think we're comfortable enough with their structure now, the next step will be to put them in their own packjage.

sckott:
okay, thanks, sounds good

PR #416, opened 2025-04-27
Draft readme for proposed v2 syntax/advice

hadley:
@sckott this is mainly for you to look at/contemplate.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@ inst/doc
 inst/ignore
 notes-redirects.R
 CLAUDE.md
+
+/.quarto/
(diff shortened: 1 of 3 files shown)


sckott:
thanks, will do

sckott (changes requested):
I like a more succinct README and point people to more resources

i think we can drop the sections

- Example packages using vcr
- Contributors

Do you think it's important someone can run the example here? Because unless they have an API key set they wouldn't be able to get the same result.

hadley:
Great, I was also thinking about removing those categories but didn't want to be too aggressive.  I think it's fine to have an example that's not runnable, but if you can think of something better I'm happy to switch.

sckott:
thanks, i'm happy with this if you are

hadley:
Yeah, it's good with me.

PR #421, opened 2025-04-28
New request matcher: compare parsed json bodies

hadley:
There's a small advantage to ignoring whitespace differences, but I think the big improvement is giving you a detailed diff if there's some small subcomponent that didn't match.


sckott:
@hadley can you resolve?

NEWS.md
```diff
@@ -17,6 +17,7 @@ vcr (development version)
 
 ## NEW FEATURES
 
+* New `body_json` request matcher that compares the parsed JSON. This both ignores differences in the textual representation of the JSON and gives more informative messages when requests don't match.
```
sckott, line 20:
```suggestion
* New `body_json` request matcher that compares the parsed JSON. This both ignores differences in the textual representation of the JSON and gives more informative messages when requests don't match. (#421)
```

sckott (changes requested):
nice, this is great

PR #425, opened 2025-04-29
Minor tweaks to logging

hadley:
I found these useful while I was doing some debugging.


R/cassette_class.R
```diff
@@ -143,27 +143,28 @@ Cassette <- R6::R6Class(
...
       self$http_interactions <- HTTPInteractionList$new(
         interactions = interactions,
         request_matchers = self$match_requests_on
       )
 
-      opts <- compact(list(
-        name = self$name,
-        record = self$record,
-        serialize_with = self$serialize_with,
-        match_requests_on = self$match_requests_on,
```
sckott, line 161: I'm not sure we want to lose what matches are being used, do we?

R/request_handler.R
```diff
@@ -24,21 +24,32 @@ RequestHandler <- R6::R6Class(
...
+        vcr_log_sprintf("  externally stubbed")
+        return(private$on_externally_stubbed_request())
       }
+
+      if (should_be_ignored(self$request)) {
+        vcr_log_sprintf("  ignored")
+        return(private$on_ignored_request())
+      }
+
+      if (cassette_active()) {
```
hadley, line 36: I needed to refactor this so we only look for the request once (instead of once to see if it's available and then a second time to actually find it).

tests/testthat/_snaps/cassette_class.md
```diff
@@ -100,3 +100,32 @@
       Error in `initialize()`:
       ! `name` must not be the same as an existing cassette.
 
+# important interactions are logged
+
+    Code
+      use_cassette("test", httr::GET(hb("/get")))
```
hadley, line 106: @sckott I think this is the best place to look at. Do you see all the information you'd want in the log?

R/cassette_class.R
sckott: got it, thanks

tests/testthat/_snaps/cassette_class.md
sckott:
```r
[Cassette: "x"] Handling request: GET http://example.com application/json, text/xml, application/xml, */* aefadsf 5
[Cassette: "x"]   looking for existing requests using uri/headers
[Cassette: "x"]   no matching requests
[Cassette: "x"] - unhandled
[Cassette: "x"]   no match: GET http://example.com
```

I assume we want to remove that `-` prefixing `unhandled`?

sckott: otherwise looks great

R/request_handler.R
sckott: looks good

tests/testthat/_snaps/cassette_class.md
hadley: How do you get that log?

sckott:
```r
library(httr)
vcr::local_vcr_configure_log()

vcr::local_cassette("test2", match_requests_on = c("method", "uri", "headers"))
GET("http://example.com")
vcr::eject_cassette()

vcr::local_cassette("test2", match_requests_on = c("method", "uri", "headers"))
GET("http://example.com", add_headers(A = 5))
vcr::eject_cassette()
```

PR #427, opened 2025-04-29
Use new checkers in `local_cassette()`

hadley:
I've aimed the code (and tests) towards `local_cassette()` since I think that will be the primary interface going forward.


sckott:
@hadley can you resolve?

sckott:
looks great, just needs resolve

hadley:
@sckott done!

PR #428, opened 2025-04-29
Create `vcr_interaction()` class

hadley:
* Check inputs
* Use `recorded_at` value when reading/writing
* Move `vcr_request()` and `vcr_response()` to the same file


sckott:
looks great

PR #429, opened 2025-04-29
Do more work in HTTPInteractionList

hadley:
Waiting on #428.

### DIFF
--- a/tests/testthat/test-http_interaction_list.R
+++ b/tests/testthat/test-http_interaction_list.R
@@ -70,3 +70,32 @@ test_that("can optionally replay", {
   interactions$response_for(req2)
   expect_equal(interactions$response_for(req2), resp2)
 })
+
+test_that("can add interactions", {
+  req1 <- vcr_request("GET", "http://a.com")
+  resp1 <- vcr_response(200, body = "a")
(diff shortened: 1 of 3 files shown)


sckott:
#428 merged

sckott:
@hadley can you resolve?

tests/testthat/test-http_interaction_list.R
```diff
@@ -70,3 +70,39 @@ test_that("can optionally replay", {
...
+  expect_equal(interactions$interactions[[1]]$response, resp1)
+  # newly added interactions can not be replayed
+  expect_equal(interactions$replayable, FALSE)
+
+  # same request replaces response
+  interactions$add(req1, resp2)
+  expect_equal(interactions$interactions[[1]]$response, resp2)
+})
+
+test_that("can add interactions", {
```
sckott, line 92: Isn't this `test_that` block an exact duplicate of the above block? maybe you copy/pasted it and meant to change it?

sckott (commented):
thanks, these changes are nice, and it's more intuitive thinking about `playable` vs. `used`

some suggestions and questions

the `i` -> `idx` is just to be consistent with the var used when using `find_request`

PR #431, opened 2025-04-29
Switch default directory to `test_path("_vcr")`

### LINKED ISSUE #395: Store fixtures in `_vcr`
What do you think about changing the default in `local_cassette()` so it uses `tests/testthat/_vcr` by default? Then you wouldn’t need a helper to change the global default and it would better match the convention used by snapshot tests.

hadley:
Fixes #395

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -17,6 +17,7 @@ vcr (development version)
 
 ## NEW FEATURES
 
+* The default path is now `tests/testthat/_vcr`. This should not affect existing packages that used `use_vcr()` because these set up a helper that sets the default directory with `vcr_configure()` (#395).
 * `local_vcr_configure()` allows you to temporarily affect vcr configuration.
 
 vcr 1.7.0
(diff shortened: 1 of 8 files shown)


sckott:
It looks like `_vcr` is the default for both `local_cassette()` and `use_cassette()` in this PR. Right? I thought in #395 we talked about only for `local_cassette()`? per the news item you added, maybe you're thinking that existing vcr users will have `vcr::vcr_configure` usage already setting the `dir` and so this change doesn't impact that usage pattern?

hadley:
Yeah, exactly. I can certainly reconsider if you want, but I thought it might be confusing to give them different defaults.

sckott:
Okay, sounds good

sckott:
wrt broken checks, maybe I resolved the conflict incorrectly? If not, i'm trying to figure out the issue here

hadley:
Not sure what went wrong there but should be better now.

sckott:
thanks!

PR #433, opened 2025-04-30
Minor httr handler refactoring

hadley:
* Use `httr::VERB()`
* Use `with::defer()`
* Improve variable name
* Use original httr request when recreating response


sckott:
@hadley can you resolve?

hadley:
Should be good now — I realised that I hadn't tested the ignored code path, and once I did that, it revealed a few more simplifications.

PR #438, opened 2025-04-30
Only skip remote tests on CRAN

hadley:


sckott:
I'm hesitant to run more tests on CRAN given that tests are run here on CI, and I'd have to deal with (most likely) rude emails - and I don't really see CRAN as a CI service.

hadley:
I think these tests should largely be safe, since they don't rely on external resources. I think it's also worth remembering that skip on CRAN tests will also be skipped when others are running revdep checks, so there's some advantage to being as minimal as possible.

sckott:
> I think these tests should largely be safe, since they don't rely on external resources

right

> skip on CRAN tests will also be skipped when others are running revdep checks

good point, hadn't thought about that

👍🏽

PR #439, opened 2025-04-30
Reduce documentation indirection

hadley:
Making it easier to assess the overall state of the documentation. I have also made a small start on a "protecting secrets" vignette as I think that's a very important topic, and shouldn't be a section of another vignette.


sckott:
The documentation files in man/rmdhunks are re-used in the http testing book https://github.com/search?q=repo%3Aropensci-books%2Fhttp-testing%20rmdhunks&type=code  so as not to have duplicated text - but I suppose we can just drop the text from man/rmdhunks into the book in those places

> I have also made a small start on a "protecting secrets" vignette as I think that's a very important topic, and shouldn't be a section of another vignette.

great, thank you!

hadley:
Yeah, in general I think that sort of documentation reuse is a mirage — in order to make a document hang together you have to rewrite to some extent to make everything flow together.

PR #442, opened 2025-04-30
Automatically compress raw bodies

### LINKED ISSUE #343: Automatically compress base64 bodies
I would be straightforward to automatically compress the body whenever it's stored as bas64. This would need a new field (maybe `base54_compressed`) but would only need a couple of lines in the serializer and the derializer.

hadley:
Introduces a new body type called "raw_gzip", ensuring that we have room to introduce new compression types in the future.

I also added more tests for body decoding/encoding so that this code is tested more directly, rather than only relying on end-to-end tests.

Fixes #343

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -17,6 +17,7 @@ vcr (development version)
 
 ## NEW FEATURES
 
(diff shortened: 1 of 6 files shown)


sckott:
Looks good thanks

PR #443, opened 2025-05-01
Don't document/export internal functions/classes

### LINKED ISSUE #374: Remove export tag from most "internals" methods?
https://docs.ropensci.org/vcr/reference/index.html#internals

I think the only ones we do need to export are ones that `webmockr` uses, which are:

- `RequestHandlerHttr2`
- `RequestHandlerHttr`
- `RequestHandlerCrul`

For these 3 keep export but make internal to keep out of docs

hadley:
I left the Cassette class exported/documented for now, since it is technically part of the exported interface of the package because it's returned by `use_cassette()` etc. I could be persauded that it should be removed too (but that should wait until after #429 to avoid a bunch of merge conflicts)

Fixes #374.


sckott:
@hadley Sorry for a meta question here. So when do you retain docs for un-exported functions? You've removed them all here. Is the idea that the code for internal functions/etc. should be written such that you don't need docs for them?

e.g, in `request_response.R` we have 

> In `response_summary()` we use [gsub] with `useBytes=TRUE` to avoid
problems sometimes seen with multibyte strings - this shouldn't affect
your data/etc. as this is only for printing a summary of the response

Which seems like the kind of thing (the "why" for a block of code) we'd want to retain.

hadley:
In general, I don't like roxygen style docs for internal functions and your example illustrates why 😄 `response_summary()` no longer uses `gsub()` but I don't expect to see that sort of internal implementation detail in the documentation so I didn't remove it. (i.e. comments far away from the source in internal functions tend to drift out of sync).

sckott:
> your example illustrates why

right, good point

PR #444, opened 2025-05-01
Deprecate `use_vcr()`

### LINKED ISSUE #423: Modify use_vcr for v2 changes

hadley:
(And re-alphabetise the imports since I seem to not know that a comes before e).

Fixes #423


sckott:
@maelle what do you think about removing `use_vcr`? I think this was your good idea in the first place.

hadley:
@maelle the reason that this is no longer necessary is that routine use of vcr no longer requires a helper. You just put `vcr::local_cassette("cassette-name")` in your test, and it automatically uses `tests/testthat/_vcr`.

maelle:
Sounds good, I opened https://github.com/ropensci-books/http-testing/issues/111

sckott:
thanks @maelle !

PR #450, opened 2025-05-01
Only record headers if needed

### LINKED ISSUE #417: Only store request headers if used?
I wonder if it makes sense to only record the request headers if they're actually used for matching. That would avoid a lot of accidental secret snafus, and would generally decrease the size of the cassettes.

hadley:
Proof of concept. Fixes #417.

@sckott started on this so you'd have something to give concrete feedback on. I do like that it eliminates the chances of leaking keys for the majority of APIs.


sckott:
I think I resolved the conflict correctly, change if not.

hadley:
Yeah, that looks reasonable to me.

sckott (commented):
We're dropping the headers slot from the interaction written to disk, but we're leaving an empty request body slot there still as `body: {}`. Maybe we should be consistent about that? Or leave body as is?

We should document this request header behavior for users. Probably in the param def for `match_requests_on` in `use_cassette()`, and maybe in the getting started vignette

sckott:
>  I do like that it eliminates the chances of leaking keys for the majority of APIs.

me too

hadley:
Yeah, let me apply the same treatment to body, and add some extra docs.

hadley:
I also need to update the secrets vignette.

hadley:
Should be good for review now.

sckott:
@hadley readme updated, look good?

hadley:
@sckott looks great!

sckott:
okay, great, merging this

PR #451, opened 2025-05-01
Make `vcr_configure()` a regular function

hadley:
So you get autocomplete when using it and we can document the parameters as arguments. The major downside is that the default rendering inside of R is not great because it's in a table and (e.g.) `clean_outdated_http_interactions` is quite wide.


sckott:
I like makimng it into a regular function. 

Should we consider renaming the longest parameter names? e..g., possibly: (with number of characters in parens)

- allow_unused_http_interactions (30) -> allow_unused_interactions (25)
- preserve_exact_body_bytes (25) ->  preserve_bytes (14)
- clean_outdated_http_interactions (32) -> clean_old_interactions (22)
- filter_sensitive_data_regex (27) -> filter_sensitive_data_r (23)

hadley:
We'd still need to deprecate the old options, but I guess we could do that via `...` and another helper function.

* `allow_unused_http_interactions` doesn't appear to be implemented? Maybe we could just drop for now?
* `preserve_exact_body_bytes` -> I was wondering if this should be `raw_body` or `body_base64`?
* I like `clean_old_interactions`, but maybe `drop_old_interactions` to be a fraction shorter?
* I'm not sure "filter" is quite the right word because it's really about replacing data. But I'm not sure how to work that in.

sckott:
- Let's drop `allow_unused_http_interactions`
- Let's do `raw_body`
- Let's do `drop_old_interactions`
- What about `swap` (4 chars) instead of `filter`? or `sub` (3 chars)?

hadley:
One more idea: how about `filter_sensitive_data` -> `hide_secrets` and then we deprecate `filter_sensitive_data_regex`. There's [only one package that uses it](https://github.com/search?q=%28org%3Aropensci+OR+org%3Acran%29+filter_sensitive_data_regex&type=code) (redbird), and we could provide a new alternative tool like `filter_body`, which would take a callback function that's applied to the body data. 

Looking at the [redbird code](https://github.com/ropensci/rebird/blob/f4e151e8c6e2bbe290b2791d733dcf152d99a1ef/tests/testthat/helper-vcr.R#L5-L9), we should also confirm that it still achieves the desired output given that it looks like it's doing a regular expression on yaml, and we no longer apply `filter_sensitive_data` directly to the on-disk representation.

sckott:
There's this project too https://github.com/CorrelAid/DataToMetabase/blob/f6a0f5c4e44b810234fa8696229ed41f8d5cf36d/tests/testthat/helper-vcr.R#L5 but i could just open an issue with suggested fix. 

I like `filter_sensitive_data` -> `hide_secrets` and deprecating `filter_sensitive_data_regex` and adding `filter_body` with a callback

> Looking at the [redbird code](https://github.com/ropensci/rebird/blob/f4e151e8c6e2bbe290b2791d733dcf152d99a1ef/tests/testthat/helper-vcr.R#L5-L9), we should also confirm that it still achieves the desired output

agreed, happy to do that

sckott:
what about the rest of the `filter_` parameters? did you like the `swap` or `sub` idea?

hadley:
Something about this idea of swapping out values when reading/writing from disk seems wrong to me, but I just can't put my finger on why. I think my gut is telling me that it's equivalent to just removing or redacting the values, so it's just more complicated for little gain.

sckott:
Okay, for the remainder of `filter_` params we can leave as is

hadley:
Now that `vcr_configure()` is a regular function, I think we can do away with the configuration vignette, which just contains largely duplicated information.

R/use_cassette.R
```diff
@@ -46,11 +46,8 @@
 #' @param allow_playback_repeats (logical) Whether or not to
 #' allow a single HTTP interaction to be played back multiple times.
 #' Default: `FALSE`.
-#' @param serialize_with (character) Which serializer to use.
-#' Valid values are "yaml" (default) and "json". Note that you can have
-#' multiple cassettes with the same name as long as they use different
-#' serializers; so if you only want one cassette for a given cassette name,
-#' make sure to not switch serializers, or clean up files you no longer need.
+#' @param serialize_with (string) Which serializer to use:
+#'   `"yaml"` (the default) or `"json"`.
```
hadley, line 50: I removed the note about uniqueness because I don't think we want to encourage people to switch serializers just to eliminate name clashes.

sckott: agree

PR #454, opened 2025-05-01
Eliminate manual editing vignette

hadley:
In favour of a comment in the getting started vignette. Let me know if you think it needs more coverage than this, but given that you can now capture errors, I think the use cases for manual editing are relatively sparse.


sckott:
right I think it should be rather rare , so this seems warranted

PR #455, opened 2025-05-01
Drop write to disk vignette

hadley:
In favour of putting the content inside the standard getting started vignette.


sckott:
thanks

PR #458, opened 2025-05-02
Record requests as we go

hadley:
I think this is a bit simpler


sckott:
thanks 👍🏽

hadley:
I just realised that the real benefit of this is that you no longer have to eject a cassette — it's still good practice, but if you miss it, there's no harm done.

PR #459, opened 2025-05-02
Simplify re-recording logic

hadley:
And move it all into one place


sckott:
Maybe we just leave as is, but it doesn't seem to make sense to use `re_record_interval` if you're not also setting `clean_outdated_http_interactions = TRUE`

And anyway is this right where where `recording: TRUE` even though in the subsequent calls the cassette file does not change? 

```r
local({
  local_cassette("test", re_record_interval = 1L)
  req <- request("https://hb.cran.dev/get")
  req_perform(req)
})
[Cassette: test] Inserting 'test.yml' (new cassette)
[Cassette: test]   recording: TRUE
[Cassette: test] Handling request: GET https://hb.cran.dev/get
[Cassette: test]   Recording response: 200 with 476 bytes of application/json data
[Cassette: test] Ejecting

local({
  local_cassette("test", re_record_interval = 1L)
  req <- request("https://hb.cran.dev/get")
  req_perform(req)
})
[Cassette: test] Inserting 'test.yml' (with 1 interactions)
[Cassette: test]   recording: TRUE
[Cassette: test] Handling request: GET https://hb.cran.dev/get
[Cassette: test]   Looking for existing requests using method/uri
[Cassette: test]     Request 1: MATCH
[Cassette: test]   Replaying response 1
[Cassette: test] Ejecting

local({
  local_cassette("test", re_record_interval = 1L)
  req <- request("https://hb.cran.dev/get")
  req_perform(req)
})
[Cassette: test] Inserting 'test.yml' (with 1 interactions)
[Cassette: test]   recording: TRUE
[Cassette: test] Handling request: GET https://hb.cran.dev/get
[Cassette: test]   Looking for existing requests using method/uri
[Cassette: test]     Request 1: MATCH
[Cassette: test]   Replaying response 1
[Cassette: test] Ejecting
```

hadley:
Doesn't look like `clean_outdated_http_interactions` is used (https://github.com/search?q=%28org%3Acran+OR+org%3Aropensci%29+clean_outdated_http_interactions&type=code&p=2), so I think we could just remove it, in favour of cleaning whenever `re_record_interval` is not `NULL`.

Yeah, I think you're right that that `recording` value is incorrect. I'll take a look later. (And figure out what's up with the tests)

sckott:
Nice, my example above works now

PR #470, opened 2025-05-05
Implement `insert_example_cassette()`

### LINKED ISSUE #309: Advice for vcr in examples?
I assume you'd want to use `inst/vcr` or similar as the path? Because of the way examples work, you'd need an explicit `insert_cassette()` and `eject_cassete()`. And then you'd put those in `\dontshow()`? So all up maybe something like this:

```R
#' \dontshow{
#'   vcr::vcr_configure(dir = system.file("vcr", package = "ellmer"))
#'   vcr::insert_cassette("content_image_url")
#' }
#' chat <- chat_openai(echo = TRUE)
#' chat$chat(
#'   "What do you see in these images?",
#'   content_image_url("https://www.r-project.org/Rlogo.png"),
#'   content_image_file(system.file("httr2.png", package = "ellmer"))
#' )
#'
#' \dontshow{dev.control('enable')}
#' plot(waiting ~ eruptions, data = faithful)
#' chat <- chat_openai(echo = TRUE)
#' chat$chat(
#'   "Describe this plot in one paragraph, as suitable for inclusion in
#'    alt-text. You should briefly describe the plot type, the axes, and
#'    2-5 major visual patterns.",
#'    content_image_plot()
#' )
#' \dontshow{vcr::eject_cassette()}
```

But then how do you force it to run the code outside of `R CMD check`? The `system.file` shim needs `devtools::load_all()` to be called.

hadley:
I thought originally that this would just be a documentation issue, but as I got deeper into it I realised that there's enough meat here to be worth a helper. Examples store cassettes in`inst/_vcr` so they're available once the package is installed.

I'm not sure how to test this, but I've set the default for `record` in such a way that the example should error during `R CMD check` if I've messed the code.

Fixes #309


sckott:
Nice! 

I tried this in a package and it is working great except it throws an empty cassette warning in the pkgdown docs like 

```
#> Warning: ✖ "wt_data" cassette ejected without recording any interactions.
#> ℹ Did you use {curl}, `download.file()`, or other unsupported tool?
#> ℹ If you are using crul/httr/httr2, are you sure you made an HTTP request?
```

I could remove the warning by adding `warn_on_empty = FALSE`  to `insert_example_cassette()`, but do we want to do that? or maybe there's something going on in pkgdown context that comes to mind?

hadley:
@sckott hmmm, I couldn't recreate that exact problem, but I found and fixed another one. If it still doesn't work for you, could you tell me exactly how you're running pkgdown?

sckott:
I''ll check with the current state and if still happening I'll make a reprex

R/examples.R
```diff
@@ -0,0 +1,98 @@
...
+  identical(Sys.getenv("IN_PKGDOWN"), "true")
+}
+
+example_cassette_path <- function(package) {
+  path <- system_file("_vcr", package = package)
+  if (path != "") {
+    path
+  } else if (is_dev_package(package)) {
+    cli::cli_inform("Creating {.path inst/_vcr}")
+    dir_create(file.path(system_file(package = package), "inst", "_vcr"))
```
sckott, line 72:
```suggestion
    sys_file <- system_file(package = package)
    path <- if (!nzchar(sys_file)) {
      file.path("inst", "_vcr")
    } else {
      file.path(normalizePath(sys_file), "inst", "_vcr")
    }
    dir_create(path)
```

Looked over this again and the path ended up being like `/inst/_vcr`, so dir created outside of the package. This change fixes it, BUT do you see any issues with this? Or make it better?

hadley:
Ok, should be better now, and made some additional changes in order to improve feedback to the user.

sckott:
Tested and seems to be working well, thanks

Issue #269, opened 2024-08-09
Context-dependent fixture recording?


sckott:
So in https://github.com/getwilds/rcromwell/issues/44 a use case has come up that it'd be nice to have context-dependent usage of cassettes. 

Consider the following use case where jobs are submitted to some engine that makes widgets. However, it takes some time before the metadata on jobs is available after submission. This is not a great situation for tests. 

```r
vcr::use_cassette("job_metadata_prep", {
  res <- submit_job(some_input = tempfile())
})

# Needed only for recording new fixture
Sys.sleep(10)

vcr::use_cassette("job_metadata", {
  res <- job_metadata(res$id)
})

expect_s3_class(res, "tbl")
```

Right now when I want to re-record the cassette (or on the very first recording I did) I have to make sure the `Sys.sleep` is not commented out because the `submit_job` call kicks off some process that leads to some wait time before appropriate data will return from `job_metadata`. (of course you may want to test these different behaviors, but lets's assume we want to test the behavior of `job_metadata` when the metadata is present. 

Consider the possibility of

```r
add_sleep <- function(sleep = 10) Sys.sleep(sleep)

vcr::use_cassette("job_metadata_prep", {
    res <- submit_job(some_input = tempfile())
  }, 
  on_new_cassette = list(after = add_sleep, before = NULL),
  if_vcr_off = list(after = add_sleep, before = NULL)
)

vcr::use_cassette("job_metadata", {
  res <- job_metadata(res$id)
})

expect_s3_class(res, "tbl")
```

Where 
- new params `on_new_cassette` and `if_vcr_off`  both accept a list of two things: before and after (or other names?), where each accepts a function, formula list or NULL, just like `httr2:::as_mock_function`
- param `on_new_cassette` is ignored if it's not a new cassette - what should happen if re-recording? if it's a new cassette then the before callback is run for before recording, and the after callback after recording

I think this takes care of the 3 scenarios in the linked issue - making it so that tests could be run without any manual toggling on/off of sleeps

hadley:
After #472, you could take an approach more like this:

```R
local({
  local_cassette("job-submit")

  res <- submit_job(some_input = tempfile())
  
  repeat {
    res <- job_metadata(res$id)
    if (res$state == "finished") {
      break
    }
    Sys.sleep(1)
  }

})
```

The challenge would be skipping the sleep in the case when you're replaying the requests. I wonder if it would be worth exposing a `cassette_state()` function, so then you could do something like:

```R
local({
[...]

PR #472, opened 2025-05-06
Always append new interactions

hadley:
I'm not sure if this is a good idea of not, but I've been contemplating the problem that repeating the same request in a cassette does not work. So in this PR, we record every request, rather than every request that doesn't match an existing request.


sckott:
I'm not sure either. Did you have a real use case from something you're working on for this, or did you just notice it by chance? Or see from #269 ?

sckott:
I checked Ruby and Python and they too record the same number of requests in the cassette as there are requests made - even of the same exact request - so the same as in this PR. I haven't checked HOW they implement that but it's a good sign that two of the siblings to this pkg behave like this PR

sckott:
> I do find it surprising that this doesn't work out of the box 

agreed

In terms of `allow_playback_repeats`, yeah I've never used it, and your search suggests it isn't used in R, but is used in [Ruby](https://github.com/search?q=language%3ARuby+allow_playback_repeats&type=code) and Python versions. I'm not sure if that's justification for keeping it or not 🤷🏽

hadley:
I added a commit removing `allow_playback_repeats`. Overall, I think this makes me more in favour of always appending, as it significantly simplifies both the code base and the explanation of how vcr works.

sckott:
I ran Jeroen's revcheck action on this branch and the results came back same as for main branch, so that was some peace of mind for me at least. 

I like this and I don't see any issues with it, and as I said above I like that this is how ruby and python versions behave (not that we have to)

ready to merge, or do you still have questions about this?

hadley:
I think we should go for it.

PR #477, opened 2025-05-09
possible fix for decode_sensitive issue

### LINKED ISSUE #475: decode_sensitive problem
I've seen this in a few different packages that use `vcr` (`rgbif`, `osfr`)

```r
Error in `vcr_request(method = request$method, uri = decode_uri(request$uri),
    body = request_body$data, headers = decode_headers(request$headers))`: `headers` must be a list, not a character vector.
```

~~Both of these packages use `crul` for http requests so perhaps its `crul` specific~~ i don't think it's `crul` specific

I think it stems from the fact that `decode_sensitive()`  - that used to be called `sensitive_put_back()` - was designed to be run on strings (see https://github.com/ropensci/vcr/blob/v1.7.0/R/serializers-yaml.R#L49), but now is run on [named lists for headers](https://github.com/ropensci/vcr/blob/main/R/serialize-headers.R#L23) but [strings for bodies](https://github.com/ropensci/vcr/blob/main/R/serialize-body.R#L40)

sckott:
- `decode_sensitive` and `encode_sensitive` were run on a named list of headers, which caused a vector to be returned instead of a named list
- changed both functions to be passed to `lapply` to run on each list element

fix #475


sckott:
A few more fixes coming here for v1 cassettes

sckott:
@hadley added a few more commits if you can take a look - may not need so many tests for `is_base64` but then again I'm not super familiar with the ins and outs of base64

R/serialize-body.R
```diff
@@ -56,12 +58,52 @@ decode_body <- function(body, preserve_bytes = FALSE) {
...
+
+  # Check if padding is valid (if present)
+  # Can only be at the end and max 2 '=' characters
+  padding_match <- grepl("^[A-Za-z0-9+/]+={0,2}$", x)
+  if (!padding_match) {
+    return(FALSE)
+  }
+
+  # Try to decode and check if it returns raw bytes
+  tryCatch(
```
hadley, line 96: IMO you could skip the rest and just do this — since this is mostly for quite old cassettes, and we're encouraging users to re-record, I don't think you need to worry about making it performant. (And probably won't be that slow anyway).

sckott: i'll try that

sckott: Changing `is_base64` to just the tryCatch block doesn't work. the problem is the following, using an example from the `rgbif` package, using the fixture at https://github.com/ropensci/rgbif/blob/master/tests/fixtures/name_parse.yml#L7, where that line has the request body that's passed in the request. here's the test https://github.com/ropensci/rgbif/blob/master/tests/testthat/test-name_parse.R#L6 and note they've set `preserve_exact_body_bytes = TRUE`

The problem is that this `jsonlite::base64_dec('["x Agropogon littoralis"]')` does return raw (doesn't fail; so passes the tryCatch), even though it's clearly not base64 encoded. That's why I had those additional checks in `is_base64` to handle these cases. 

preserve bytes in v1 I'm pretty sure was only applied to response bodies, so now in v2 code that handles preserve bytes is being used in request bodies when I don't think any v1 request bodies will be base64 encoded, though maybe there are?

I don't think it needs to be performant, but I don't think we should make cran checks fail for any revdeps

sckott: Yeah I'm not sure why `jsonlite::base64_dec` doesn't  catch all the base64 rules 🤷🏽

PR #480, opened 2025-05-12
Polish logging

hadley:
* Drop quotes from around cassette name. I don't think these are generally needed since cassette names are file names.
* Log file name instead of `serialize_with`.
* Log the consequence of `record_mode`, i.e. whether or not the cassette is actually recording
* Only log that we're looking for requests if there are some to find
* During the matching process, label the request with its index (since you can look it up in the cassette), and generally improve the visual hierarchy.


sckott:
LGTM

PR #484, opened 2025-05-13
Show simple error if logging turned on

hadley:
I don't think there's a need for additional details since you see it all during the matching process.


sckott:
I like it. Thinking about possibly removing `UnhandledHTTPRequestError`

PR #485, opened 2025-05-13
chore: typo fix

maelle:


DESCRIPTION
```diff
@@ -61,7 +61,7 @@ Config/testthat/start-first: ause_cassette_re_record
 Encoding: UTF-8
 Language: en-US
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.3.2
+RoxygenNote: 7.3.2.9000
```
sckott, line 64: @maelle ROR support requires dev `roxygen2` I assume, but do we know when the next verison of `roxygen2` will hit cran?

sckott (commented):
thanks @maelle , one question for you

DESCRIPTION
maelle: No I don't know, but I could remove this change from the PR.

sckott: Yeah, let's go back to the current cran version of `roxygen2` for now, and I've opened an issue to upgrade when `roxygen2` is updated on cran https://github.com/ropensci/vcr/issues/486

maelle: Done!

sckott: thanks!

PR #487, opened 2025-05-14
Minor logging tweaks

hadley:
* Use upper case "Recording" to match other top-level results
* Convert URL params to a list to get a better diff from waldo
* Indent diff an extra level


sckott:
LGTM

PR #491, opened 2025-05-14
Drop external stubbing

hadley:
Since it was never used, is very isolated change, and I think it's clearer now that vcr doesn't have to adhere to the exact interface of the ruby package.


sckott:
yeah I don't think we need to adher to ruby's vcr. +1 for removing

PR #495, opened 2025-05-14
Drop crayon in favour of cli

hadley:
I included a basic test of `verbose = TRUE` just to double check I didn't screw anything up.


sckott (commented):
Don't we want to remove `crayon` from the DESCRIPTION file?

hadley:
Yes 😬

PR #498, opened 2025-05-15
Revdep checks from Posit infrastructure

hadley:


sckott:
Thanks for using posit infra to run checks. 

Is the idea to merge this now, or iterate on checks until all are green?

hadley:
Merge it now?

sckott:
will merge then

PR #499, opened 2025-05-15
cassette_state fxn to get current cassette state: recording, replaying or NULL

sckott:
#269 

First pass at this, not saying I've thought super deeply about this

to do:
- [ ] add tests

questions
- are there other states besides recording, replaying?
- should absence of recording/replaying be `NULL` or something else? raise an error since you shouldn't be calling the function if a cassette is not active?

So this now works well:

```r
library(vcr)
library(httr2)
library(tictoc)

submit_job <- function() {
  cat("job submitted\n")
}
job_metadata <- function(...) {
  request("https://hb.cran.dev/anything") %>% 
    req_body_json(list(number = sample.int(100, 1))) %>% 
    req_perform() %>% 
    resp_body_json() %>%
    .$json %>% 
    .$number
}

tic("recording")
local({
  local_cassette("job-submit2")

  res <- submit_job()
  
  repeat {
    number <- job_metadata(res$id)
    if (number < 10) {
      break
    }
    if (cassette_state() != "replaying") {
      Sys.sleep(1)
    }
  }
})
toc()
#> recording: 47.344 sec elapsed

tic("replaying")
local({
  local_cassette("job-submit2")

  res <- submit_job()
  
  repeat {
    number <- job_metadata(res$id)
    if (number < 10) {
      break
    }
    if (cassette_state() != "replaying") {
      Sys.sleep(1)
    }
  }
})
toc()
#> replaying: 0.293 sec elapsed
```


hadley:
There's the problem of `new_episodes` where you are both recording and replaying.  I wonder if it makes sense to start by adding a `replaying` boolean to match the existing `recording` field/

sckott:
okay, lets merge #459 first

Issue #500, opened 2025-05-19
Error when posting images in vcr  v1.7.0.91: `rawToChar(b)`: embedded nul in string


christopherkenny:
Hello. This is a great package, so first, thank you!

I'm converting `bskyr` from testing using `httptest2` to use `vcr`, as there are several endpoints which require making multiple HTTP requests to perform an action, such as making a Bluesky post.

However, whenever I include an image in the request, it fails with some form of the error:

```
Error in `rawToChar(x$body$data)`: embedded nul in string
```

Here's an offending, in-context example:

```r
library(vcr)
library(testthat)
library(bskyr)

auth <- bs_auth(user = get_bluesky_user(), pass = get_bluesky_pass())

test_that('`bs_new_embed_external()` works', {
  vcr::local_cassette('e_external', preserve_exact_body_bytes = TRUE)
  embed <- bs_new_embed_external(
    uri = 'https://christophertkenny.com/bskyr/'
  )
  expect_vector(embed, vector(mode = 'list'))
})
```
which results in the following error and trace:

```r
── Error: `bs_new_embed_external()` works ─────────────────────────────────────────────────────────────────────────────────────
[...]

PR #503, opened 2025-05-29
httr2 image body handling refinement

sckott:
This works but wouldn't work if a request that had an image body (or similar) didn't have an appropriate header - which I imagine is rare at least

#500


hadley:
@sckott I hope you don't mind my hijacking your PR, but my commit should make it more clear that we assume binary, UNLESS it's clearly marked as text. That makes this change much safer overall.

sckott:
@hadley all good. that's a good change, just needed to flip the logic.

PR #516, opened 2025-06-25
Polish configure docs

hadley:
And remove unnused `ignore_request` argument

Not perfect, but better than what's there (which is missing some words)


sckott:
thanks!

PR #517, opened 2025-06-26
Match linguist regex

hadley:
So that cassettes appear as generated files in GitHub PR diffs. This means that they're not shown by default, but can be expanded if desired.


hadley:
@sckott I think you could argue this either way, but this is what happens with the original ruby VCR package, and if you don't hide them, you can get A LOT of output to look at it (e.g. https://github.com/tidyverse/ellmer/pull/591/files)

PR #520, opened 2025-06-27
More easily expose cassette status outside of vcr

hadley:
I think this might end up being a nicer way to disable env var checks when you're replaying a cassette.


R/use_cassette.R
```diff
@@ -290,3 +295,24 @@ check_cassette_name <- function(x, call = caller_env()) {
 
   invisible()
 }
+
+
+#' Determine if vcr is recording/replaying
+#'
+#' [local_cassette()] and `with_cassette()` set the `VCR_IS_RECORDING`
+#' and `VCR_IS_REPLAYING` environment variables to make it easy to determine
+#' vcr state without to take a dependency on vcr. These functions show you
```
sckott, line 304:
```suggestion
#' vcr state without taking a dependency on vcr. These functions show you
```

hadley:
This is what I'm thinking of for ellmer:

```R
key_get <- function(name, error_call = caller_env()) {
  val <- Sys.getenv(name)
  if (!identical(val, "")) {
    val
  } else {
    if (is_replaying()) {
      # this should never get used, but we don't want to error
      ""
    } else if (is_testing()) {
      testthat::skip(sprintf("%s env var is not configured", name))
    } else {
      cli::cli_abort("Can't find env var {.code {name}}.", call = error_call)
    }
  }
}
```

I think it fits pretty naturally in this sort of helper.

sckott:
looks like a good approach, maybe we could even provide a helper fxn for folks - though maybe there's too many scenarios to generalize it

PR #525, opened 2025-06-30
Reuse chunk label as cassette name

### LINKED ISSUE #512: `cassette: true`
Support `cassette: true` to mean use the same name as the label.

hadley:
With `cassette: true`. Fixes #512


hadley:
Ooops no, I was just supposed to make it error temporarily so I could look at the error that knitr throws.

sckott:
thanks

PR #528, opened 2025-07-08
Remove webmockr

### LINKED ISSUE #388: Remove dependency on stubbing requests within webmockr?
We're tightly integrated with `webmockr` for now, where the hooks live that hook into the 3 http libraries, which then shuttle data back to `vcr` when `vcr` is in use. 

What would it look like to disconnect these two so that `vcr` works without `webmockr`?

sckott:
fix #388 

Note that I haven't reconsidered code brought over from webmockr in detail - so there may be more room for improvement/simplifying/etc 

There's still some fxns in webmockr that have similar names, e.,g., `httr_mock`, so we prefix here with vcr just to avoid confusion 

Associated PRs (no need to look at these unless you want to)
- https://github.com/ropensci/crul/pull/183 
- https://github.com/ropensci/webmockr/pull/150


hadley (approved):
I didn't review the code in detail, but this looks like a great simplification!

PR #545, opened 2025-07-22
Skip test on CRAN

hadley:
`expect_snapshot()` is not run on CRAN so this cassette is never ejected, causing problems for all subsequent tests that use a cassette with the same name (which is a lot of them!)

You can verify the fix by running `devtools::check(env_vars = c(NOT_CRAN = "false"))`.

### DIFF
--- a/tests/testthat/test-lightswitch.R
+++ b/tests/testthat/test-lightswitch.R
@@ -8,14 +8,13 @@ test_that("can turn off and turn on", {
   expect_true(turned_on())
 })
 
-test_that("turned_off works iif no cassettes active", {
-  insert_cassette("test", warn_on_empty = FALSE)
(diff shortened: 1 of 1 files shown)


sckott:
With those last few commits, this passes checks on winbuilder - I don't think I've pinpointed EXACTLY what made it work, but at least this is a minimal set of changes

Tests are failing here b/c I removed `Remotes: ropensci/crul`

sckott:
- Reverted the 3 commits I made
- Added that change to the lightswitch test from slack

Tests pass with `NOT_CRAN=false` with crul v1.6 in release and devel R for me - but continue to fail with crul v1.5 - I think Kurt is okay with vcr failing with crul v1.5

PR #548, opened 2025-08-20
feat: simplify use of vcr in examples

### LINKED ISSUE #546: roxygen2 tag
How nice would it be to have something like

```r
#' @exampleVcr cassette-name
#' <code>
```

instead of

```r
#' @examples 
#' \dontshow{insert_example_cassette("cassette-name", package = "package-name")}
#' <code>
#' \dontshow{eject_example_cassette()}
```

Now, the only usage of `insert_example_cassette()` I could find is in ellmer where there is a bit more setup/cleaning anyway: https://github.com/tidyverse/ellmer/blob/4223c73094a7fb4027f811283fa30c73182c921b/R/utils.R#L263 @hadley 

Beside, `insert_example_cassette()` has other arguments. Maybe the syntax of the examples could somehow be closer to the syntax of vignette chunks, so that the arguments could be passed as knitr options? 

In any case, the syntax for vcr usage in examples feels very clunky compared to the smoothness of using vcr in tests and vignettes. 😇

maelle:
Fix #546

Example usage: https://github.com/EMODnet/emodnet.wfs/pull/193


R/roxygen-example.R
```diff
@@ -0,0 +1,7 @@
+#' @rdname vcr-example
```
maelle, line 1: I'd like to add an actual test but am not sure how to.

R/roxygen.R
```diff
@@ -0,0 +1,124 @@
...
+#'
+#' This functions amends DESCRIPTION to set up vcr's roxygen2 machinery.
+#' Call it once when you're setting a new package.
+#'
+#' @param path character path to the package
+#'
+#' @return Nothing: called for file system side effects.
+#' @export
+#'
+use_vcr_examples <- function(path = ".") {
```
maelle, line 70: From https://github.com/moodymudskipper/devtag/blob/main/R/use-devtag.R

R/examples.R
```diff
@@ -2,7 +2,11 @@
 #'
 #' @description
```
maelle, line 3: In this file it'd be useful to describe _how_ to run the examples. `devtools::run_examples()`? Use `example()`? Run the code by opening the Rd file? Which reminds me of https://github.com/posit-dev/positron/issues/1734

R/roxygen.R
```diff
@@ -0,0 +1,124 @@
...
+#' @importFrom roxygen2 roxy_tag_rd
+NULL
+
+#' @export
+roxy_tag_parse.roxy_tag_examplesVCR <- function(x) {
+  roxygen2::tag_markdown(x)
+}
+
+#' @export
+roxy_tag_rd.roxy_tag_examplesVCR <- function(x, base_path, env) {
```
maelle, line 11: I looked at https://github.com/r-lib/roxygen2/pull/993/files

DESCRIPTION
```diff
@@ -33,6 +33,7 @@ Imports:
     lifecycle,
     R6,
     rlang (>= 1.1.0),
+    roxygen2,
```
maelle, line 36: I'd have preferred to make this an optional dependency but I don't know how to not have the imports like roxy_tag_parse in DESCRIPTION.

Should the tag/roclet actually live in yet another package? So `use_vcr_examples()` would set up a dependency on that package, called vcr.roxygen2 or something like that?

R/examples.R
drmowinckels: Yes! I ended up running my examples line by line, which was a little tedious.

maelle:
I wonder how useful this would actually be, given that there might be other boilerplate code anyway: in ellmer an option is set and unset, in @drmowinckels' package a function is called to fake authentication. 

Could one somehow use `@examples` with a code chunk like in a vignette, with "eval" (=examplesIf) and "cassette" options for instance? And a knitr hook would allow having the right boilerplate code with dontshow.

Could one also store some boilerplate for _all examples_ in a central place, like a general setup chunk for all examples, that could set local options for instance?
Or for that should package maintainers write their own `@examples` tag overwriting the roxygen2 one?

I might be overthinking this. :sweat_smile:

R/roxygen.R
```diff
@@ -0,0 +1,124 @@
+#' @importFrom roxygen2 roxy_tag_parse
+#' @importFrom roxygen2 roxy_tag_rd
```
hadley, line 2: I think you could drop these and switch from `@export`ing the method to `@S3method` registering it so you don't need to import roxygen2.

hadley: Sorry, I meant `@exportS3Method`, https://roxygen2.r-lib.org/articles/namespace.html#s3

maelle:
I simply expect you as a roxygen2 maintainer to not break this precious tag in the future :pleading_face:

maelle:
Sort of related: https://github.com/r-lib/roxygen2/issues/1697

maelle:
@mpadge I'm sorry but I didn't understand.

I thought you meant moving the body of `roxy_tag_parse.roxy_tag_examplesVCR()` to `roclet_process.roclet_examplesVCR()` which didn't work. 

Any chance you could do a PR to this PR? :sweat_smile: Thanks in advance.

R/examples.R
maelle: @sckott see this, we're actually not sure whether there's an easier way. Not sure how it was done in ellmer by @hadley.

PR #557, opened 2026-06-19
potential fix for delete fxn when type = tests

sckott:
- if `cassette_path` gives `NULL` abort with message
- add small test to cover this use case

Thoughts?

The way this came up was that I was in a package of mine that uses `vcr` and I loaded vcr then tried to run `delete_cassettes("some-pattern")` and did not get a helpful error:

```r
Error in `dir.exists()`:
! invalid filename argument
```

Hopefully this makes sense?

This I think only matters when trying to delete cassettes in tests directories

### DIFF
--- a/R/delete_cassettes.R
+++ b/R/delete_cassettes.R
@@ -78,17 +78,7 @@ delete_cassettes <- function(
 delete_type_cassettes <- function(type, prefix) {
   dir <- switch(
     type,
-    tests = {
-      path <- cassette_path()
-      if (is.null(path)) {
-        cli::cli_abort(c(
-          "No cassette directory has been configured.",
-          i = "Call {.fn vcr_configure} or {.fn local_vcr_configure} with a {.arg dir} argument before using {.fn delete_cassettes}.",
-          i = "Example: {.code vcr_configure(dir = \"tests/testthat/_vcr\")}"
-        ))
-      }
-      path
-    },
+    tests = test_cassette_dir(),
     examples = file.path(rprojroot::find_package_root_file(), "inst", "_vcr"),
     vignettes = file.path(
       rprojroot::find_package_root_file(),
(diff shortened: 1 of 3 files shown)


R/delete_cassettes.R
```diff
@@ -78,7 +78,17 @@ delete_cassettes <- function(
...
   dir <- switch(
     type,
-    tests = cassette_path(),
+    tests = {
+      path <- cassette_path()
+      if (is.null(path)) {
+        cli::cli_abort(c(
+          "No cassette directory has been configured.",
+          i = "Call {.fn vcr_configure} or {.fn local_vcr_configure} with a {.arg dir} argument before using {.fn delete_cassettes}.",
+          i = "Example: {.code vcr_configure(dir = \"tests/testthat/_vcr\")}"
```
sckott: @maelle i'm not sure about this message. thoughts? what should we suggest users do?

R/delete_cassettes.R
```diff
@@ -78,7 +78,17 @@ delete_cassettes <- function(
 delete_type_cassettes <- function(type, prefix) {
   dir <- switch(
     type,
-    tests = cassette_path(),
+    tests = {
```
maelle: Maybe factor this out in a function? It feels weird not to see the whole `switch()` call at once.

R/delete_cassettes.R
maelle: `"tests/testthat/_vcr"` is the default. Does vcr find cassettes when they are there? If so should the example path be `"custom-path/"`? If not shouldn't vcr find cassettes in the default directory?

maelle (commented):
Thanks and sorry for the delay!

sckott:
Thanks @maelle  , happy belated Bastille day! Having a look at your comments now.

R/delete_cassettes.R
sckott: done

R/delete_cassettes.R
sckott: @maelle Well, yes, "tests/testthat/_vcr" is the default location where cassettes are placed, BUT `vcr_configure()` doesn't have a default for the `dir` argument, so the user has to pass something to it. I think when in a flow of adding/running tests, it's all done automatically, but that's not true for deleting cassettes. Do you think we should have vcr auto-discover cassettes for this delete cassettes workflow?

maelle (approved):
Thank you!

PR #558, opened 2026-07-18
Changes for modern httr2

### LINKED ISSUE #532: Bump httr2 requirement and remove old code
In a year or so, we can remove `modern_httr2()` and code in the `FALSE` branches.

sckott:
fixes #532 

The one thing I'm not sure about this https://github.com/ropensci/vcr/blob/main/R/serialize-headers.R#L48 We still need the function but should we change for modern httr2?

### DIFF
--- a/R/request_handler-httr2.R
+++ b/R/request_handler-httr2.R
@@ -4,6 +4,7 @@ RequestHandlerHttr2 <- R6::R6Class(
 
   public = list(
     initialize = function(request) {
+      check_httr2_version()
       if (!length(request$method)) {
         request$method <- httr2_method(request)
       }
@@ -68,6 +69,19 @@ RequestHandlerHttr2 <- R6::R6Class(
     }
   )
 )
+
+check_httr2_version <- function(call = rlang::caller_env()) {
+  if (!is_installed("httr2", version = "1.2.0")) {
+    cli::cli_abort(
+      c(
+        "vcr requires {.pkg httr2} >= 1.2.0.",
+        "i" = "Installed version: {.val {as.character(utils::packageVersion('httr2'))}}.",
+        "i" = "Please update httr2 with {.code install.packages(\"httr2\")}."
+      ),
+      call = call
+    )
+  }
+}
 
 httr2_method <- function(req) {
   getNamespace("httr2")$req_get_method(req)
--- a/R/request_handler-httr2.R
+++ b/R/request_handler-httr2.R
@@ -6,14 +6,14 @@ RequestHandlerHttr2 <- R6::R6Class(
     initialize = function(request) {
       check_installed("httr2", version = "1.2.0")
       if (!length(request$method)) {
-        request$method <- httr2_method(request)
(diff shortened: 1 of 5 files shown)


R/request_handler-httr2.R
```diff
@@ -69,38 +70,30 @@ RequestHandlerHttr2 <- R6::R6Class(
   )
 )
 
-httr2_method <- function(req) {
-  if (modern_httr2()) {
-    return(getNamespace("httr2")$req_get_method(req))
+check_httr2_version <- function(call = rlang::caller_env()) {
+  if (!is_installed("httr2", version = "1.2.0")) {
```
hadley: You can use `check_required()` here

R/request_handler-httr2.R
```diff
@@ -69,38 +70,30 @@ RequestHandlerHttr2 <- R6::R6Class(
...
-    req$method
-  } else if ("nobody" %in% names(req$options)) {
-    "HEAD"
-  } else if (!is.null(req$body)) {
-    "POST"
-  } else {
-    "GET"
-  }
+httr2_method <- function(req) {
+  getNamespace("httr2")$req_get_method(req)
```
hadley: You should be able to drop `getNamespace()` etc now. That was only necessarily previously to avoid an `R CMD check` warning when modern httr2 wasn't yet available.

tests/testthat/test-request_handler-httr2.R
```diff
@@ -1,3 +1,26 @@
+test_that("works with the installed (modern) httr2", {
+  skip_if_not_installed("httr2", "1.2.0")
+  expect_no_error(check_httr2_version())
+
+  local_vcr_configure(dir = withr::local_tempdir())
+  req <- httr2::request(hb("/get"))
+  use_cassette("modern_httr2", resp <- httr2::req_perform(req))
+  expect_s3_class(resp, "httr2_response")
+  expect_equal(resp$status_code, 200)
+})
```
hadley: I don't think you need an explicit test for this; it's implicit in all the other httr2 tests.

hadley:
Why do we still need `request_headers_redact()`?

R/request_handler-httr2.R
sckott: Thanks, will do

R/request_handler-httr2.R
sckott: I assume you meant `check_installed` here?

sckott:
> Why do we still need `request_headers_redact()`?

What i'm proposing is to change the function to (or similar):

```r
request_headers_redact <- function(headers) {
  matches <- match("authorization", tolower(names(headers)))
  matches <- matches[!is.na(matches)]

  if (length(matches) == 0) {
    headers
  } else {
    headers[-matches]
  }
}
```

Because we still want to automatically remove Authorization headers across all http clients, yes?

hadley:
Oh gotcha. If it's used for other backends, then you'll need to keep it.

PR #559, opened 2026-08-03
Up minimum required R version to match `{yaml12}`

PietrH:
I recently had a Github Actions run fail on a windows R4.1 machine for my package `{etn}` because `{yaml12}` requires R4.2. By adopting this dependency all packages dependent on vcr now also require R4.2 as a minimum. 

In this PR I just document this change in DESCRIPTION.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -25,7 +25,7 @@ URL: https://github.com/ropensci/vcr/,
     https://docs.ropensci.org/vcr/
 BugReports: https://github.com/ropensci/vcr/issues
 Depends:
-    R (>= 4.1)
+    R (>= 4.2)
 Imports:
     cli,
     curl (>= 6.3.0),


sckott:
Thanks @PietrH . So would it be a problem for you to require R 4.2 or greater? I'm wondering if I should undo this change - and put `yaml` back just so pkgs that depend on vcr don't have to change their minimum R. Thoughts?

PietrH:
> Thanks @PietrH . So would it be a problem for you to require R 4.2 or greater? I'm wondering if I should undo this change - and put `yaml` back just so pkgs that depend on vcr don't have to change their minimum R. Thoughts?

Not particularly, I try to keep my minimum R requirement as low as possible. Other than that I have no real objections apart from just documenting it in vcr. 

Is vcr actually using any of the new yaml12 functionality at the moment?

sckott:
@PietrH 

> Is vcr actually using any of the new yaml12 functionality at the moment?

Just the drop in replacements for `yaml` - the goal was to use a more reliable, and presumably faster alternative to the old `yaml` package.

PietrH:
Hi Scott, what do you think is the way forward. Should downstream dependencies update to R4.2+, or should vcr offer a yaml12 alternative for lower R versions?

sckott:
@PietrH I reverted the change to yaml12, we're back to yaml, i didn't want to make folks using vcr change. So closing this. Thanks for opening this! and looking forward to any issues or contributions in the future

